PR #6, opened 2016-04-05
Import from CSV file

macejmic:
closes https://github.com/3scale/3scale_toolbox/issues/7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+.DS_Store
--- a/exe/3scale-import
+++ b/exe/3scale-import
@@ -1,0 +1,178 @@
+#!/usr/bin/env ruby
+
+require '3scale_toolbox/cli'
+require 'optparse'
+require '3scale/api'
+require 'uri'
+require 'csv'
+
+options = {}
+
+parser = OptionParser.new do |parser|
+  parser.banner = '3scale import <command> [options]'
+
+  parser.on('-d', '--destination DESTINATION') do |domain|
+    options[:destination] = domain
+  end
+
+  parser.on('-f', '--file FILE') do |file|
+    options[:file] = file
+  end
+
+  parser.on('-h', '--help', 'Prints this help') do
+    puts parser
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -7,3 +7,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+.DS_Store
```
mikz: this really should be in your global gitignore :) but i'm with it it here too

examples/import.md
```diff
@@ -0,0 +1,33 @@
+## Import from a CSV file
```
mikz: this should not be in `examples` folder no ? but rather `docs` ?

exe/3scale-import
```diff
@@ -0,0 +1,178 @@
...
+
+  services.keys.each do |service_name|
+    # create service
+    service = client.create_service name: service_name
+
+    if service['errors'].nil?
+      stats[:services] += 1
+      puts "Service #{service_name} has been created."
+    else
+      raise "Service has not been saved. Errors: #{service['errors']}"
```
mikz: might be better to `abort`. there is no need for stacktrace in the cli app

mikz:
Like it! I think it is totally fine for now.

.gitignore
macejmic: done :)

docs/import_example.csv
```diff
@@ -0,0 +1,7 @@
+service_name,endpoint_name,endpoint_http_method,endpoint_path,auth_mode,endpoint_system_name,type
```
mikz: this one can be in `examples` folder :)

andrewdavidmackenzie:
+1 for including example files.

For this toolbox, do we need to plan some "self-documentation" (help option? man page? other?) for people who will install the gem from a gem server?

mikz:
@andrewdavidmackenzie the plan is that it should be self describing. Run the command without anything , get help. Run one of those without params, get help. Etc. Like any other CLI tool there.

Man page makes sense also. But I'd say it is not a priority for now.

PR #28, opened 2017-02-20
Overwrite service

mpguerra:
Adds a new command 3scale update to copy changes from one existing service to another

### DIFF
--- a/exe/3scale-update
+++ b/exe/3scale-update
@@ -1,0 +1,222 @@
+#!/usr/bin/env ruby
+
+require '3scale_toolbox/cli'
+
+require 'optparse'
+require 'pry'
+
+options = {}
+
+parser = OptionParser.new do |parser|
+  parser.banner = '3scale update <command> [options]'
+
+  parser.on('-s', '--source SOURCE', "Source") do |domain|
+    options[:source] = domain
+  end
+
+  parser.on('-d', '--destination DESTINATION', "Destination") do |domain|
+    options[:destination] = domain
+  end
+
+  parser.on('-t', '--target-system-name TARGET_SYSTEM_NAME', 'Target system name') do |name|
+    options[:target_system_name] = name
+  end
+
+  parser.on('-h', '--help', 'Prints this help') do
+    puts parser
+    puts
+    puts 'Available Commands:', ['service <service_id>', 'help']
+    exit
+  end
+end
+
+print_help = ->(error = nil) do
+  if error
+    puts "Error: #{error}"
+    puts
+  end
+  parser.parse(['--help'])
+end
+
+
+parser.parse!
+
+def compare_hashes(first, second, keys)
+  keys.map{ |key| first.fetch(key) } == keys.map{ |key| second.fetch(key) }
+end
+
+def fetch_option(options, key)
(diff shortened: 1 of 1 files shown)


mpguerra:
Companion to 3scale/3scale-api-ruby#20

exe/3scale-update
```diff
@@ -0,0 +1,215 @@
...
+        client.create_application_plan_limit(target_id, metrics_mapping.fetch(limit.fetch('metric_id')), limit)
+      end
+      puts "target application plan #{target_id} is missing #{missing_limits.size} from the source plan #{source_id}"
+    end
+  when 'help'
+    print_help.call
+  when nil
+    print_help.call("missing subcommand")
+  else
+    print_help.call("unknown command #{command}")
```
mikz: probably should `exit 1` after to return proper exit code, the same missing subcommand

exe/3scale-update
```diff
@@ -0,0 +1,215 @@
...
+def compare_hashes(first, second, keys)
+  keys.map{ |key| first.fetch(key) } == keys.map{ |key| second.fetch(key) }
+end
+
+def fetch_option(options, key)
+  options.fetch(key) { raise OptionParser::MissingArgument, key }
+end
+
+def provider_key_from_url(url)
+  url[/\w*@/][0..-2]
```
mikz: Any reason why not to use https://github.com/3scale/3scale_toolbox/blob/master/exe/3scale-import#L44-L53 ?

mpguerra: nope, was not aware that one was different :)

mayorova:
I've added a bit of structure to see the different steps of the copy process better.
Also, 95% of these steps are exactly the same as used in the `copy` command, so I think it could be worth extracting the logic somewhere (now it's `ServiceUpdater` class, but can be renamed/moved).

There is an option added to only copy the mapping rules (`--rules-only`), and to optionally delete the existing mapping rules on target before copying the ones from source (`-f`).

exe/3scale-update
```diff
@@ -0,0 +1,270 @@
...
+    uri      = URI(url)
+    uri.user = nil
+    
+    uri.to_s
+  end
+
+  def target_service_params(source)
+    # NOTE: backend_version and deployment_option are not yet returned by show_service method
+    params = %w(name backend_version deployment_option end_user_registration_required)
+    Hash[params.collect {|p| [p, source[p]]}].reject { |key, value| !value }
```
mikz: Might be better to use `Array#to_h` instead of `Hash[]` just for consistency with other methods.

mikz: Also might be easer just to: 

```ruby
source.select { |k,v| params.include?(k) && v }
```

exe/3scale-update
```diff
@@ -0,0 +1,270 @@
...
+      end
+    end
+
+    puts "updating limits for application plans..."
+
+    application_plan_mapping = target_client.list_service_application_plans(target_service_id).map do |plan_target|
+      plan = source_plans.find{|plan| plan.fetch('system_name') == plan_target.fetch('system_name') }
+
+      [plan['id'], plan_target['id']]
+    end
```
mikz: I think this does not work. It does not call `.to_h` so it is just array of tuples and then the next each expects it to be hash

Issue #49, opened 2018-07-02
Login command to manage remotes


eguzki:
Command to 
- Add remote (currently *provider key* and provider *URI*)
- List remotes
- Show remote info
- Remove remote
- Rename remote

andrewdavidmackenzie:
I propose "Add a remote" checks that the credentials provided are correct, and doesn't add it if it can't confirm them.

PR #53, opened 2018-07-20
3scale remotes 

eguzki:
implementation of #49


mikz:
@eguzki 👍 for opening an early PR.

My suggestion for the data store would be [PStore](http://ruby-doc.org/stdlib-2.5.1/libdoc/pstore/rdoc/PStore.html) or [YAML::Store](http://ruby-doc.org/stdlib-2.5.1/libdoc/yaml/rdoc/yaml/Store.html) It is not so well know, but amazing piece of work. Both are in standard library and have simple API to use. 

I used it in https://github.com/3scale/jenkins-build/blob/master/lib/jenkins/build/configuration.rb and it works great.

lib/3scale_toolbox.rb
```diff
@@ -1,9 +1,32 @@
+require '3scale_toolbox/configuration'
+
 module ThreeScaleToolbox
+  @config_file = nil
```
mikz: Is this really needed to be module level? 

I imagine it is possible to provide an instance of `ThreeScaleToolbox::Configuration` to the `--config-file` option. Then there would be no need to "remember" the file that was passed in parameters and modify global state (good luck in tests). 

And every command would have it available as an option.

lib/3scale_toolbox.rb
```diff
@@ -1,9 +1,32 @@
...
     plugin_paths.each { |plugin_path| require plugin_path }
   end
 
   def self.plugin_paths
     Gem.find_files('3scale_toolbox_plugin')
   end
+
+  def self.config_file
+    # 3SCALERC env var has priority over $HOME/.3scalerc.yaml file
+    @config_file ||= ENV['THREESCALERC']
```
mikz: `THREESCALERC` is a pretty weird name. Why not something more meaningful like `THREESCALE_CLI_CONFIG' ?

lib/3scale_toolbox/commands/3scale_command.rb
```diff
@@ -12,6 +12,9 @@ def self.command
           usage       '3scale <command> [options]'
           summary     '3scale CLI Toolbox'
           description '3scale CLI tools to manage your API from the terminal.'
+          required :c, 'config-file', '3scale CLI configuration file' do |val, _cmd|
```
mikz: Passing config file is required?

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -19,9 +19,16 @@ def self.command
           end
         end
 
+        # list remotes
         def run
-          # list remotes
-          puts "list remotes"
+          if ThreeScaleToolbox.configuration.remotes.empty?
+            puts 'Emtpy remote list.'
+            exit 0
```
mikz: Calling `exit 0` from commands will make it quite hard to test. IMO we should rely on other mechanisms rather than just directly exiting.

lib/3scale_toolbox/configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+module ThreeScaleToolbox
+  class Configuration
+    ATTRIBUTES = %i[remotes].freeze
+    private_constant :ATTRIBUTES
+
+    def initialize
+      @store = YAML::Store.new(ThreeScaleToolbox.config_file)
+    end
+
+    def remotes
```
mikz: This is not very scalable. Every command that wants to store something in its own namespace will need to extend this object.

IMO we should look for better API where each command can put own stuff into the configuration object.

lib/3scale_toolbox/configuration.rb
```diff
@@ -0,0 +1,45 @@
+require 'yaml/store'
+
+module ThreeScaleToolbox
+  class Configuration
+    ATTRIBUTES = %i[remotes].freeze
+    private_constant :ATTRIBUTES
+
+    def initialize
+      @store = YAML::Store.new(ThreeScaleToolbox.config_file)
```
mikz: IMO this is wrong direction of dependencies. The file name of the store should be pushed here, not pulled from somewhere.

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
```diff
@@ -17,7 +17,37 @@ def self.command
...
+        end
+
+        def validate_remote_new_name(name)
+          remotes = ThreeScaleToolbox.configuration.remotes
+          raise "Could not rename, new name '#{name}' already exists." if remotes.key? name
+        end
+
+        def rename_remote(remote_old_name, remote_new_name)
+          validate_remote_old_name remote_old_name
+          validate_remote_new_name remote_new_name
```
mikz: Why this uses ruby equivalent of global variable? You can't inject the configuration instance and it is loaded from a module.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
```diff
@@ -17,7 +17,36 @@ def self.command
...
+          begin
+            add_remote(*arguments[0..1])
+          rescue StandardError => e
+            warn e.message
+            exit 1
+          end
+        end
+
+        def validate_input_params
+          return unless arguments.size != 2
```
mikz: This is not done by the CLI library? You can't require number of arguments ? It still executes the command?

eguzki:
testing is definitely needed. Will implement some.

lib/3scale_toolbox/commands/3scale_command.rb
eguzki: no. When `--config-file` is in the command line, then some value is required.

mikz: Yep, I found it in the CRI docs. Quite poor naming decision.

lib/3scale_toolbox/commands/remote_command.rb
eguzki: Yeah, it will be abstracted. Console IO will be abstracted as well  to be easy to test.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
eguzki: not sure, will check.

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
eguzki: Configuration has setter method and can be injected. Very convenient for testing.

mikz: It is not injected. It is read from a global variable. Thats quite a difference. It would be injected if it would be part of these messages or object instance.

lib/3scale_toolbox.rb
eguzki: Would be nice to "inject" configuration object as part of options available to all commands. Do not know how, though.

Config file has setter method and can be injected. Then calling configuration object for the first time, that config file will be used.

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
eguzki: Would be nice to "inject" configuration object as part of options available to all commands. Do not know how, though. Has to be "read" using global variable at some point.

lib/3scale_toolbox.rb
mikz: After reading cri source and docs I'm not impressed. There there is no builtin support transformations of options/arguments or validations. 
The only thing I found is `option_added` delegate to do some transformations. IMO we could use that to transform the configuration file name into the configuration object, that would be passed to every command in options. Then every command would get own instance of this object based on the CLI flag / env / default value.
But the "delegate" feature looks very internal and not really usable from the outside.

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
mikz: The object responsible for instantiating the runner is `cri`. It will parse command line options, instantiate the runner and execute it. That one should be responsible for passing the configuration in the options when instantiating the runner. Then every instance will get its own without the need for global variable. 
Unfortunately it looks like cri does not really support any argument transformation / validation 
https://github.com/3scale/3scale_toolbox/pull/53#discussion_r205044182 like for example clamp has: https://github.com/mdub/clamp#parsing-and-validation-of-options-and-parameters.

eguzki: I mostly agree. 

I am going to open issue for feature request about option validation/transformation as well as arguments.

However, do not going to change CLI framework for now. Let's finish remotes feature and bring some other functionality to have something up and running. Then, CLI framework feature lacking issue can be raised.

eguzki: https://github.com/ddfreyne/cri/issues/67

lib/3scale_toolbox/configuration.rb
eguzki: I do not like the idea of exposing a hash object so each command can put/get/delete own stuff.  Having class level `ATTRIBUTES` list enables filtering what's can be included in persisted file. 

Instead of making commands extend this object with methods like 
```ruby
def remotes
end 
```
and
```ruby
def update_remotes
end 
```
those methods could be generated dynamically just adding new elements to `ATTRIBUTES` list. The body of the methods should be the same.

mikz: If I write command as and extension I would not be able to use the config file without monkey patch. That does not sound right to me. In the and it is all Hash structures.

There should be some good reason to write more code than necessary. I really don't see one here. The configuration class can just make sure it opens transaction when running the cli, read / write data and thats it.

I really don't see any benefit of `configuration.remotes` over `configuration.data['remotes']`.

Why would you need to filter / limit what is included in the file? What is the problem we would be solving with this technical solution?

eguzki: I guess this has to do with the typical trade-off flexibility over control. We allow plugins to add stuff into persisted config file and lose control over what is stored vs plugins are not allowed to add stuff (can read it though) and stored content is controlled (at least at first level).

Rubygems does not allow arbitrary content in the config file. All updates are via write accessors like [set_api_key](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/config_file.rb#L302)

mikz: I don't really see it as a tradeoff. I see it as that there should be some problem to be solved and if there is no problem then there is no reason to write code.

What is the problem we are trying to solve with restricting what can be stored in the config?
That is a serious question. Is it namespaces between commands ?

BTW Rubygems offers raw access: https://github.com/rubygems/rubygems/blob/672ddcbe14f49c48f6c5492b38480a9162680b5b/lib/rubygems/config_file.rb#L447-L455

eguzki: Right, Rubygems offers raw access. 

I guess control over stored content allows you to provide business logic over it, in case needed. 

In your own example, you add control over stored content.
https://github.com/3scale/jenkins-build/blob/master/lib/jenkins/build/configuration.rb

mikz: My example was done as one day hack to solve a problem and never aimed to be extensible.
It stored configuration just for itself, not for any number of commands written by other people.

If providing some "nicer" access is the issue then the command can implement that by itself  using OpenStruct or something. The business logic on the data can be implemented anywhere.

The responsibility logic of the store should be implemented in the store: transaction, load, save, ...

For me the core question is: will other commands store configuration there?

If the answer is yes, then imagine the situation of 100 commands storing the configuration there. The work that would have to be done to make each of them work.
Then offer the simplest access that fulfills the goal.

If the answer is no, then offer the simplest way for the app to read its config and thats it. No need to be extensible.

Because ultimately there should be just single responsibility. Either being extensible configuration store for data and each command will operate on it how they want or a one instance of this: store for one application with business logic.

You can serialize whatever ruby objects in the YAML store. Each configuration object can have own small config with business logic object stored and serialized.

eguzki:
configuration object with api to add freely any stuff

lib/3scale_toolbox/commands/3scale_command.rb
eguzki: :+1:  will update definition of commands for clearness

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
eguzki: doing some refactor using CLI framework new features

lib/3scale_toolbox/commands/remote_command/remote_add.rb
eguzki: done by  CLI library

eguzki:
Some work has been done simplifying code using new features of CLI library. Still unittests and integration tests missing.

lib/3scale_toolbox/commands/copy_command/copy_service.rb
```diff
@@ -4,83 +4,70 @@
...
-          exit 1
-        end
-
-        def self.fetch_required_option(options, key)
-          options.fetch(key) { exit_with_message "error: Missing argument #{key}" }
+        def run
+          source      = fetch_required_option(:source)
+          destination = fetch_required_option(:destination)
+          system_name = fetch_required_option(:target_system_name)
+          copy_service(arguments[:service_id], source, destination, system_name)
```
mikz: This is really a nitpick, but I think a whitespace before this line would be really useful.

lib/3scale_toolbox/commands/import_command/import_csv.rb
```diff
@@ -163,11 +152,10 @@ def self.import_csv(destination, file_path, insecure)
...
 
-        def self.run(opts, _)
-          destination = fetch_required_option(opts, :destination)
-          file_path = fetch_required_option(opts, :file)
-          insecure = opts[:insecure] || false
-          import_csv(destination, file_path, insecure)
+        def run
+          destination = fetch_required_option(:destination)
+          file_path = fetch_required_option(:file)
+          import_csv(destination, file_path)
```
mikz: The same comment in some other command: IMO declaring variables and running the actual action would be nice to be separated by a newline.

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -0,0 +1,61 @@
...
+            usage       'remote <command> [options]'
+            summary     '3scale CLI remote'
+            description '3scale CLI command to manage your remotes'
+            runner RemoteCommand
+          end
+        end
+
+        def run
+          list_remotes
+        rescue StandardError => e
```
mikz: Do you have RuboCop/CodeClimate set up? 

There is no need for explicitly catch `StandardError` as `rescue => e` does the same and requires less thinking for the reader to figure out what it actually does.

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -0,0 +1,61 @@
...
+          warn e.message
```
mikz: I think it is nicer to just pass the whole error to the warning, so the error itself can customize how it is going to be serialized: `warn e`.

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -0,0 +1,61 @@
...
+          exit 1
+        end
+
+        def invalid_remote
+          raise "invalid remote configuration from config file #{config_file}"
+        end
+
+        def validate_remotes(remotes)
+          return if remotes.nil?
+          invalid_remote unless remotes.class == Hash
```
mikz: This looks pretty much like a switch statement already.


```ruby
case remotes
when nil then return
when Hash then remotes.each_value(&method(:validate_remote))
else invalid_remote
end
```

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -0,0 +1,61 @@
...
+          remotes.each_value do |remote|
+            invalid_remote unless remote.key?(:endpoint) && remote.key?(:provider_key)
+          end
+        end
+
+        def list_remotes
+          remotes = config.data :remotes
+
+          validate_remotes(remotes)
```
mikz: I think the exception should not be thrown from inside the validation, but rather from here.


```ruby
return invalid_remote unless valid_remotes?
```

And `valid_remotes?` would just return `true/false`. There is no need to raise exceptions unless in exceptional situation.

lib/3scale_toolbox/commands/remote_command.rb
```diff
@@ -0,0 +1,61 @@
...
+
+          if remotes.nil? || remotes.empty?
+            puts 'Emtpy remote list.'
```
mikz: Emtpy 👀

lib/3scale_toolbox/commands/remote_command/remote_add.rb
```diff
@@ -0,0 +1,75 @@
...
+          # 'arguments' cannot be converted to Hash
+          add_remote arguments[:remote_name], arguments[:remote_url]
+        rescue StandardError => e
+          warn e.message
+          warn e.backtrace
+          exit 1
+        end
+
+        def validate_remote_name(name)
+          remotes = config.data :remotes
```
mikz: Wouldn't it be better to actually return an empty Hash instead of nil? That would reduce the defensive programming you have to do in few places.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
```diff
@@ -0,0 +1,75 @@
...
+          raise 'fatal: remote name already exists.' if !remotes.nil? && remotes.key?(name)
+        end
+
+        def parse_remote_uri(remote_url_str)
+          # should raise error on invalid urls
+          remote_uri_obj = URI(remote_url_str)
+          provider_key = remote_uri_obj.user
+          remote_uri_obj.user = ''
+          endpoint = remote_uri_obj.to_s
+          { provider_key: provider_key, endpoint: endpoint }
```
mikz: It is not necessarily a provider_key. I'd just call it authentication.

lib/3scale_toolbox/commands/remote_command/remote_rename.rb
```diff
@@ -0,0 +1,52 @@
...
+          # 'arguments' cannot be converted to Hash
+          rename_remote arguments[:remote_old_name], arguments[:remote_new_name]
+        rescue StandardError => e
+          warn e.message
+          exit 1
+        end
+
+        def validate_remote_old_name(name)
+          remotes = config.data :remotes
+          raise "Could not rename, old name '#{name}' does not exist." unless !remotes.nil? && remotes.key?(name)
```
mikz: It is always a bad idea to use `unless ... && ...` for readability reasons. Also this condition is used all over the place and IMO it can be cleaned up. Either by abstracting the whole "remote store" to own object or by some utility functions.

lib/3scale_toolbox/configuration.rb
```diff
@@ -0,0 +1,38 @@
...
+      @store_data = nil
+      @store.transaction do
+        @store[key] = yield @store[key]
+      end
+    end
+
+    private
+
+    def store_data
+      @store_data ||= read
```
mikz: This makes it not threadsafe. IMO it is better to just open transaction for every read and keep the store to do its' job.

lib/3scale_toolbox/commands/remote_command.rb
eguzki: RuboCop is my friend

```bash
$ rubocop -v
0.59.2
```
Rubocop is not warning, with no specific conf, all settings are default.

Anyway, error handling will be done globally and not by command, so that code is being removed.

lib/3scale_toolbox/commands/remote_command.rb
eguzki: this error handling per command will be deleted

lib/3scale_toolbox/commands/remote_command/remote_add.rb
eguzki: will call it `auth_key` because it is a key or token, not a method or process.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
eguzki: And enforce Configuration values to be Hash objects?

mikz: I personally prefer Structs or own classes, but that works too.

eguzki:
I think remotes commands implementation is done. Last review and acceptance is missing, of course.

Still missing integration of remotes with other commands, like copying services. But I guess it deserves other PR.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
eguzki: I have implemented `ThreeScaleToolbox::Remotes` module to abstract all operations regarding remotes. It includes some validations and parsing logic, besides accessing/updating remote data from configuration.

WDYT?

lib/3scale_toolbox/commands/remote_command/remote_remove.rb
```diff
@@ -25,9 +28,8 @@ def run
         private
 
         def remove_remote(remote_name)
-          config.update(:remotes) do |remotes|
-            remotes = {} if remotes.nil?
-            remotes.tap do |r|
+          update_remotes do |rmts|
+            rmts.tap do |r|
```
mikz: Why `rmts` over `remotes` ?

lib/3scale_toolbox/remotes.rb
```diff
@@ -0,0 +1,68 @@
...
+    module ClassMethods
+    end
+
+    ##
+    # Fetch remote list
+    # Perform validation
+    #
+    def remotes
+      rmts = config.data :remotes
+      return invalid_remote unless validate_remotes?(rmts)
```
mikz: I think this would be better as `raise invalid_remote` as this `return` implies it should return something, but in fact it never does.

lib/3scale_toolbox/commands/remote_command/remote_add.rb
mikz: That's better! I personally prefer classes over modules, but it is still an improvement 👍

lib/3scale_toolbox/commands/remote_command/remote_remove.rb
eguzki: trying not to shadow `remotes` method from `Remotes` module

mikz: Is there some rubocop check for that? I don't see why. Shadowing a method by a local variable is pretty common in Ruby.

eguzki: nope, just common sense :)

PR #55, opened 2018-07-30
Testing plugin framework

eguzki:
Basic testing infrastructure and tests for plugin framework
- [x] plugin detection tests
- [x] command/subcommand hierarchy build tests


spec/unit/command_hierarchy_spec.rb
```diff
@@ -0,0 +1,57 @@
+require '3scale_toolbox/cli'
+
+# `expect(...).not_to matcher.and matcher` is not supported
+RSpec::Matchers.define_negated_matcher :not_raise_error, :raise_error
```
mikz: Of course. Because every line is this assertion. Raising exception fails the test.

spec/unit/command_hierarchy_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    10.times.each do |cmd_idx|
+      expect do
+        ThreeScaleToolbox::CLI.run(["cmd_#{cmd_idx}"])
+      end.to not_raise_error.and output("command cmd_#{cmd_idx}\n").to_stdout
+    end
+  end
+
+  it 'vertical tree loaded with add_subcommand' do
```
mikz: This is seriously unreadable. Honestly I'd not even bother trying to understand this kind of test.

It overrides variables (`cmd`), defines dynamically content that can be handrolled, does assertion on dynamic data. All of those are test smells in my book.

spec/unit/plugin_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+require '3scale_toolbox/cli'
+
+RSpec::Matchers.define_negated_matcher :not_raise_error, :raise_error
+
+RSpec.describe 'Plugin command' do
+  include_context :temp_dir
+  include_context :plugin
+  include_context :random_name
+
+  before(:each) do
```
mikz: Why not `around` ?

spec/unit/command_hierarchy_spec.rb
eguzki: When your code can raise `SystemExit`, it is common to add matcher just to make sure `exit` is not raised. If raised, error is catched, other tests are run and everything works as expected. If matcher not added, `SystemExit` makes rspec process to end and many tests are not run.

All tests running
```
ThreeScaleToolbox::CLI.run([name])
```

must include `assert_error` or negative to avoid problems.

spec/unit/plugin_spec.rb
eguzki: it's exactly the same

mikz: Maybe implementation wise, but definitely not readability wise. 
Around tells the reader it is one logical operation done around the test. This is two independent operations.

spec/unit/command_hierarchy_spec.rb
mikz: So you want to test that your code calls or does not call `exit` by rescuing all exceptions? 

Wouldn't it be better just to globally rescue `SystemError` in around block and fail the test?

Then any test that does not assert that it exits would fail. And you don't need negative assertions everywhere - as testing library designers intended.

spec/unit/3scale_toolbox_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'fileutils'
```
mikz: What do you need from FileUtils ? I see only references to File.

spec/unit/plugin_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  around(:each) do |example|
+    $LOAD_PATH.unshift(tmp_dir) unless $LOAD_PATH.include?(tmp_dir)
+    example.run
+    $LOAD_PATH.delete(tmp_dir)
+  end
+
+  it 'is not loaded when not in load path' do
+    expect do
+      ThreeScaleToolbox::CLI.run(%w[simple])
+    end.to raise_error.and output(/unknown command/).to_stderr
```
mikz: Is it expected to exit? If so, then I guess you should have a matcher that says something `exit_with(1)` to be explicit it exits with some status code.

spec/unit/3scale_toolbox_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  let(:name) { random_lowercase_name }
+  let(:dest_plugin_file) { File.join(tmp_dir, '3scale_toolbox_plugin.rb') }
+
+  before(:each) do
+    plugin = get_plugin_content(name.capitalize, name)
+    File.open(dest_plugin_file, 'w') do |file|
+      file.write(plugin)
+    end
+    $LOAD_PATH.unshift(tmp_dir) unless $LOAD_PATH.include?(tmp_dir)
```
mikz: I guess this should be in the `temp_dir` context. You are using it in both tests anyway.

spec/shared_contexts.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.shared_context :temp_dir do
+  around(:each) do |example|
+    Dir.mktmpdir('3scale_toolbox_rspec-') do |dir|
+      @tmp_dir = dir
+      example.run
+    end
+  end
+
+  attr_reader :tmp_dir
```
mikz: Why not using `let` ?

spec/spec_helper.rb
```diff
@@ -0,0 +1,2 @@
+RSpec.configure do |config|
```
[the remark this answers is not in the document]
eguzki: no, I wil let rspec generate it.

spec/unit/3scale_toolbox_spec.rb
eguzki: old stuff, will be deleted

spec/unit/3scale_toolbox_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  include_context :plugin
+  include_context :random_name
+
+  let(:name) { random_lowercase_name }
+  let(:dest_plugin_file) { File.join(tmp_dir, '3scale_toolbox_plugin.rb') }
+
+  before(:each) do
+    plugin = get_plugin_content(name.capitalize, name)
+    File.open(dest_plugin_file, 'w') do |file|
+      file.write(plugin)
```
mikz: You can just use `File.write(dest_plugin_file, plugin)`.

spec/unit/3scale_toolbox_spec.rb
eguzki: separation of concerns. Temporary directories might be useful not only for adding to load path.

eguzki:
something else pending @mikz ?

spec/unit/command_hierarchy_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      define_singleton_method :command do
+        Cri::Command.define do
+          name        command_name
+          usage       command_name
+          runner      this_class
+        end
+      end
+
+      define_method :run do
+        puts "command #{command_name}"
```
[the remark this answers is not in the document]
eguzki: The test is about expecting a given command to be executed. There are several commands and each command does a different output on execution. If a param is passed on execution, any command could be executed with the same output and no clue about which command was executed.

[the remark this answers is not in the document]
eguzki: Ok, let me rewrite the test, maybe easier to understand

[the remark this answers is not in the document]
mikz: But you can pass in the message when creating those commands in the test.

```ruby
a = create_command('a', 'Hi, Im command A')
b = create_command('b', 'this is different command')
c = create_command('c', 'third is the charm')
```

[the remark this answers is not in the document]
eguzki: I did not get what you meant. Doing this slight modification.

spec/unit/command_hierarchy_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        command_name: random_lowercase_name,
+        message: 'Lorem ipsum dolor sit amet'
+      },
+      {
+        command_name: random_lowercase_name,
+        message: 'One upon a time'
+      }
+    ]
+
+    commands.each do |command|
```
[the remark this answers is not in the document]
mikz: `.each` for me makes only sense in the test if it is used to generate test cases, not to verify assertions inside the test

Also random data are necessary only in fuzzing tests, as they decrease readability and you can't refer to the values when they fail.

I've seen a lot of tests and making something that is clear to everyone even after a year is quite hard task, but definitely worth it. It should be so simple, that person that has no familiarity with the project should understand what it does.

Also the test should have reasonable message. Here it is `it('.add_command').` That is not valid message for me. 
`.add_command` is a context, not a test. It should be something like: `it('adds commands that can be later executed')` if that is what it is testing.

eguzki:
Then, we need to workout testing culture and patterns and get closer into some common, acceptable, tests development basics

mikz:
http://www.betterspecs.org is a good resource.

PR #61, opened 2018-10-19
3scale_toolbox.gemspec: remove bundler devel dep

eguzki:
[Ruby 2.3 by Software Collections](https://www.softwarecollections.org/en/scls/rhscl/rh-ruby23/) includes bundler which version (1.10) conflicts with required bundler version (1.11) right now. 

http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-ruby23/

Removing this requirement to enable RPM packaging for RHEL 7 with Ruby 2.3 enabled by Software Collections


mikz:
Why not just requiring 'any' bundler version? It is there so the release tasks work.

Issue #32, opened 2017-11-29
complete duplication of a service


lucamaf:
the following settings are not copied when copying a service:
the base path, integration mode and API authentication are not copied to the new service.

Ideally the tool would be used to have a full duplicate and then just modify the staging and production endpoints.

eguzki:
Opened issue on *porta* related to this issue:
https://github.com/3scale/porta/issues/133

eguzki:
https://github.com/3scale/3scale_toolbox/pull/64 fixes about 
* integration mode (deployment_mode)
* API authentication (backend_version)

Still missing base path, proxy settings. WIP

eguzki:
Base path, proxy settings and proxy settings are being copied. 

I guess all attributes you requested are being copied. Can you verify @lucamaf ?

Try HEAD from upstream:

Ruby >= 2.3 required
```
$ git clone git@github.com:3scale/3scale_toolbox.git
$ cd 3scale_toolbox
$ gem build 3scale_toolbox.gemspec
$ gem install 3scale_toolbox-0.4.0.gem
```

PR #64, opened 2018-10-23
include new available attributes on copy/update service

eguzki:
New required attributes of the service are available now, https://github.com/3scale/porta/issues/133.

Add them on copy/update service commands. 

Does not fix entirely https://github.com/3scale/3scale_toolbox/issues/32, but this is an improvement.

### DIFF
--- a/lib/3scale_toolbox/commands/copy_command/copy_service.rb
+++ b/lib/3scale_toolbox/commands/copy_command/copy_service.rb
@@ -54,13 +54,6 @@ module ThreeScaleToolbox
           url.sub /\w*@/, ''
         end
 
-        def self.service_valid_params
-          %w[
-            name backend_version deployment_option description
-            system_name end_user_registration_required
-            support_email tech_support_email admin_support_email
-          ]
-        end
 
         # Returns new hash object with not nil valid params
         def self.filter_params(valid_params, source)
@@ -70,7 +63,7 @@ module ThreeScaleToolbox
         end
 
         def self.copy_service_params(original, system_name)
-          service_params = filter_params(service_valid_params, original)
+          service_params = filter_params(Commands.service_valid_params, original)
           service_params.tap do |hash|
             hash['system_name'] = system_name if system_name
           end
--- a/lib/3scale_toolbox/commands/copy_command/copy_service.rb
+++ b/lib/3scale_toolbox/commands/copy_command/copy_service.rb
@@ -54,13 +54,26 @@ module ThreeScaleToolbox
           url.sub /\w*@/, ''
         end
 
+        def self.service_valid_params
+          %w[
+            name backend_version deployment_option description
+            system_name end_user_registration_required
+            support_email tech_support_email admin_support_email
+          ]
+        end
+
+        # Returns new hash object with not nil valid params
+        def self.filter_params(valid_params, source)
+          valid_params.each_with_object({}) do |key, target|
+            target[key] = source[key] unless source[key].nil?
+          end
+        end
+
         def self.copy_service_params(original, system_name)
-          {
-            name:        original['name'],
-            system_name: system_name,
-            backend_version: original['backend_version'],
-            end_user_registration_required: original['end_user_registration_required']
-          }.reject { |key, value| !value }
+          service_params = filter_params(service_valid_params, original)
+          service_params.tap do |hash|
+            hash['system_name'] = system_name if system_name
+          end
         end
 
         def self.copy_service(service_id, source, destination, system_name, insecure)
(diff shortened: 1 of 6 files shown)


lib/3scale_toolbox/commands/copy_command/copy_service.rb
```diff
@@ -54,13 +54,26 @@ def self.endpoint_from_url(url)
...
+            name backend_version deployment_option description
+            system_name end_user_registration_required
+            support_email tech_support_email admin_support_email
+          ]
+        end
+
+        # Returns new hash object with not nil valid params
+        def self.filter_params(valid_params, source)
+          valid_params.each_with_object({}) do |key, target|
+            target[key] = source[key] unless source[key].nil?
```
mikz: What about:

```ruby
target[key] = source[key] if source.key?(key)
```

lib/3scale_toolbox/commands/update_command/update_service.rb
```diff
@@ -67,8 +67,11 @@ def endpoint_from_url(url)
           end
 
           def target_service_params(source)
-            # NOTE: backend_version and deployment_option are not yet returned by show_service method
-            params = %w(name backend_version deployment_option end_user_registration_required)
+            params = %w[
```
mikz: Why this does not reuse `service_valid_params` somehow?

lib/3scale_toolbox/commands/update_command/update_service.rb
```diff
@@ -67,8 +67,11 @@ def endpoint_from_url(url)
...
+              name backend_version deployment_option description
+              system_name end_user_registration_required
+              support_email tech_support_email admin_support_email
```
mikz: This is not tested, right?

spec/unit/copy_service_spec.rb
```diff
@@ -31,4 +33,48 @@
...
+      end
+    end
+
+    it 'all expected params are copied' do
+      new_service_params = described_class.copy_service_params(source_service_params, nil)
+
+      expect(new_service_params).to include(*params)
+    end
+
+    it 'extra params are not copied' do
```
mikz: Why we don't want to copy all params?

lib/3scale_toolbox/commands/copy_command/copy_service.rb
eguzki: key needs to exist and the value not nil

spec/unit/copy_service_spec.rb
eguzki: trying to avoid sending unexpected params to 3scale API. Just for sanity.

eguzki: this is the response body from `GET /admin/api/services/#{id}`
```
{
  "id": 2555417761220,
  "name": "some_name_1",
  "state": "incomplete",
  "system_name": "some_name_1",
  "end_user_registration_required": true,
  "backend_version": "2",
  "deployment_option": "hosted",
  "support_email": "thomas@webtypes.com",
  "description": "some description",
  "created_at": "2018-10-17T11:12:46+02:00",
  "updated_at": "2018-10-17T11:12:46+02:00",
  "links": [
    {
      "rel": "metrics",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220/metrics"
    },
    {
      "rel": "end_user_plans",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220/end_user_plans"
    },
    {
      "rel": "self",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220"
    },
    {
      "rel": "service_plans",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220/service_plans"
    },
    {
      "rel": "application_plans",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220/application_plans"
    },
    {
      "rel": "features",
      "href": "https://3scale-supertest-admin.3scale.net/admin/api/services/2555417761220/features"
    }
  ]
}
```

3scale API to create service `POST /admin/api/services.xml`

does not need many of them. 

Instead you would allow them all to be included in create service or update service request body?

mikz: Hmmm. I'm not sure if the API is going to just strip unacceptable parameters or throw an error.
If it would strip them, then we are free to send it whole back. Then there would be no need for maintenance when we expose more fields. Or going for a blacklist instead of a whitelist.

But anyway, it is good as it is.

eguzki: the idea was whitelist as it is :+1:

eguzki:
* Added tests for `update service` command, specifically, the part that filters service parameters.
* Reuse valid service params array

PR #68, opened 2018-10-31
update service: only include system_name when specified in options

eguzki:
When updating service from source to destination, source `system_name` is being sent to update destination `system_name`. This is a problem when updating services within the same tenant, since you would get the  following error from API:
```json
{"system_name": ["has already been taken"]}
```
The fix is: do not try to overwrite system name, unless specified using command's new optional argument with option `-t, --target_system_name`

### DIFF
--- a/lib/3scale_toolbox/commands/update_command/update_service.rb
+++ b/lib/3scale_toolbox/commands/update_command/update_service.rb
@@ -15,6 +15,7 @@ module ThreeScaleToolbox
 
             required  :s, :source, '3scale source instance. Format: "http[s]://<provider_key>@3scale_url"'
             required  :d, :destination, '3scale target instance. Format: "http[s]://<provider_key>@3scale_url"'
+            option    :t, 'target_system_name', 'Target system name', argument: :required
             flag      :f, :force, 'Overwrites the mapping rules by deleting all rules from target service first'
             flag      :r, 'rules-only', 'Updates only the mapping rules'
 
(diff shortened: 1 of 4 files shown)


lib/3scale_toolbox/commands/update_command/update_service.rb
```diff
@@ -34,9 +35,9 @@ def self.fetch_required_option(options, key)
         end
 
         class ServiceUpdater
-          attr_reader :source_client, :target_client, :source_service_id, :target_service_id
+          attr_reader :source_client, :target_client, :source_service_id, :target_service_id, :target_system_name
 
-          def initialize (source, source_service_id, destination, target_service_id, insecure)
+          def initialize (source, source_service_id, destination, target_service_id, insecure, system_name)
```
mikz: This starts to be a bit too much.

lib/3scale_toolbox/commands/update_command/update_service.rb
```diff
@@ -66,8 +68,16 @@ def endpoint_from_url(url)
             uri.to_s
           end
 
+          # system name only included when specified from options
           def target_service_params(source)
-            source.select { |k, v| Commands.service_valid_params.include?(k) && v }
+            target_service = source.select do |k, v|
+              Commands.service_valid_params.include?(k) \
```
mikz: Wouldn't it be nicer to provider `service_valid_update_params` that does not have the `system_name` ?

lib/3scale_toolbox/commands/update_command/update_service.rb
eguzki: Sure! 
refactor is coming with new features of the CLI framework :)

lib/3scale_toolbox/commands/update_command/update_service.rb
eguzki: I was thinking on that too. But `system_name` is needed sometimes, and it is a valid attribute of the service.

PR #70, opened 2018-11-06
(WIP) Command refactor

eguzki:
copy service, update service and import csv commands need refactoring. Very long methods, code duplicated and not idiomatic ruby code.

So far, only copy service has been refactored.

### DIFF
--- a/lib/3scale_toolbox/commands/copy_command/copy_service.rb
+++ b/lib/3scale_toolbox/commands/copy_command/copy_service.rb
@@ -1,4 +1,5 @@
 require 'cri'
+require '3scale/api'
 require '3scale_toolbox/base_command'
 
 module ThreeScaleToolbox
@@ -6,6 +7,20 @@ module ThreeScaleToolbox
     module CopyCommand
       class CopyServiceSubcommand < Cri::CommandRunner
         include ThreeScaleToolbox::Command
+        include ThreeScaleToolbox::Remotes
+
+        attr_reader :source, :destination, :system_name, :service_id,
+                    :source_remote, :copy_remote,
+                    :source_service, :copy_service,
+                    :source_proxy,
+                    :source_metrics, :copy_metrics,
+                    :source_methods, :copy_methods,
+                    :source_hits, :copy_hits,
+                    :source_plans, :copy_plans,
+                    :source_mapping_rules, :copy_mapping_rules,
+                    :metrics_mapping,
+                    :application_plan_mapping
+
         def self.command
           Cri::Command.define do
             name        'service'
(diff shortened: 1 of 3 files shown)


lib/3scale_toolbox/commands/copy_command/copy_service.rb
```diff
@@ -1,11 +1,26 @@
...
 module ThreeScaleToolbox
   module Commands
     module CopyCommand
       class CopyServiceSubcommand < Cri::CommandRunner
         include ThreeScaleToolbox::Command
+        include ThreeScaleToolbox::Remotes
+
+        attr_reader :source, :destination, :system_name, :service_id,
+                    :source_remote, :copy_remote,
+                    :source_service, :copy_service,
```
mikz: This looks a lot like `source` and `copy` are the same kind of object.

If they are, it would be nicer to have just `source` and `destination` and methods `metrics` and others implemented there.

You know, it is a *lot* of instance variables. I'd try to not have so many and rather group them into more meaningful objects.

lib/3scale_toolbox/commands/copy_command/copy_service.rb
```diff
@@ -23,183 +38,263 @@ def self.command
...
 
-        def compare_hashes(first, second, keys)
-          keys.map{ |key| first.fetch(key) } == keys.map{ |key| second.fetch(key) }
-        end
+        private
 
-        def provider_key_from_url(url)
-          url[/\w*@/][0..-2]
+        def create_remotes
+          @source_remote = get_remote source
```
mikz: Is this method "creating" remotes? It just gets them.

Also, I don't really like methods that just set instance variables.
And I personally use instance variables only in the initializer. Everything else has to be done through accessors.

lib/3scale_toolbox/commands/copy_command/copy_service.rb
```diff
@@ -23,183 +38,263 @@ def self.command
...
         end
+
+        def process
+          create_remotes
+          fetch_source_service
+          create_service
+          fetch_source_proxy_settings
+          create_proxy_settings
+          fetch_source_metrics_methods
+          fetch_copy_metrics_methods
```
mikz: I think this is insane.

There is an undescribed order between these 18 method calls. Also, because all of them mutate the object's state somehow, it is really mind-boggling to see when is what available. Some of the methods prepare data for later commands, some actually do external actions.

IMO this can be refactored to several smaller classes that each do its part and make the state more immutable by defining the source/copy objects and loading all necessary information upfront.

eguzki:
Mainly I agree. I did not like it either, but it was a job done and wanted other opinion.

PR #73, opened 2018-11-09
Global error handler

eguzki:
Main error handler for all commands

- print just error when error type is expected or generated in a controlled manner
- print verbose information when unexpected error happen. In addition to stack trace dump, `crash.log` is created with app version, system information, installed gems and load paths

### DIFF
--- a/exe/3scale
+++ b/exe/3scale
@@ -4,4 +4,4 @@ require '3scale_toolbox'
 
 args = ARGV.clone
 
-exit(ThreeScaleToolbox::CLI.run(args) ? 1 : 0)
+exit(ThreeScaleToolbox::CLI.run(args))
--- a/lib/3scale_toolbox/cli.rb
+++ b/lib/3scale_toolbox/cli.rb
@@ -33,10 +33,11 @@ module ThreeScaleToolbox::CLI
 
   def self.run(args)
     install_signal_handlers
-    ErrorHandler.error_watchdog do
+    err = ErrorHandler.error_watchdog do
       load_builtin_commands
       ThreeScaleToolbox.load_plugins
       root_command.build_command.run args
     end
+    err.nil? ? 0 : 1
   end
 end
--- a/lib/3scale_toolbox/cli/error_handler.rb
+++ b/lib/3scale_toolbox/cli/error_handler.rb
@@ -9,15 +9,13 @@ module ThreeScaleToolbox
 
       # Catches errors and prints nice diagnostic messages
       def error_watchdog
-        error = false
-        begin
-          # Run
-          yield
-        rescue StandardError, ScriptError => e
-          handle_error e
-          error = true
-        end
-        error
+        # Run
+        yield
+      rescue StandardError, ScriptError => e
+        handle_error e
+        e
+      else
+        nil
       end
 
       private
--- a/spec/unit/error_handler_spec.rb
+++ b/spec/unit/error_handler_spec.rb
@@ -22,10 +22,10 @@ RSpec.describe ThreeScaleToolbox::CLI::ErrorHandler do
         end
       end
 
-      it 'returns true' do
+      it 'returns error' do
         expect(
           subject.error_watchdog { raise_toolbox_error }
-        ).to be_truthy
+        ).to be
       end
     end
 
@@ -39,16 +39,16 @@ RSpec.describe ThreeScaleToolbox::CLI::ErrorHandler do
         end
       end
 
-      it 'returns true' do
+      it 'returns error' do
         expect(
           subject.error_watchdog { raise_runtime_error }
-        ).to be_truthy
+        ).to be
       end
     end
 
     context 'Does not raise error' do
       it 'returns true' do
-        expect(subject.error_watchdog {}).to be_falsey
+        expect(subject.error_watchdog {}).to be_nil
       end
     end
   end


lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+  # Catches errors and prints nice diagnostic messages
+  class ErrorHandler
+    def self.error_watchdog(exit_on_error: true)
+      new.error_watchdog(exit_on_error: exit_on_error) { yield }
+    end
+
+    def error_watchdog(exit_on_error:)
+      # Set exit handler
+      %w[INT TERM].each do |signal|
+        Signal.trap(signal) do
```
mikz: Signal.trap is global, so I'd rather seen it in the main CLI module.

lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def write_version_information(stream)
+      write_section_header(stream, 'Version Information')
+      stream.puts ThreeScaleToolbox::VERSION
+    end
+
+    def write_system_information(stream)
+      write_section_header(stream, 'System Information')
+      stream.puts `uname -a`
```
mikz: You can use `Etc.uname` for that.

lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+        true
+      else
+        false
+      end
+    end
+
+    def print_error(error)
+      write_error(error, $stderr)
+
+      File.open('crash.log', 'w') do |io|
```
mikz: I think it is better to just print it to the STDERR.

For example, in containers, the filesystem can disappear and you are only left with stdout/stderr.

lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+    private
+
+    def handle_error(error, exit_on_error:)
+      if expected_error?(error)
+        warn
+        warn "Error: #{error.message}"
+      else
+        print_error(error)
+      end
+      exit(1) if exit_on_error
```
mikz: Seems like the better place for this is the main CLI module. This flag is being passed across several methods in the end to just call exit.

lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,133 @@
...
+      end
+
+      # Set stack trace dump handler
+      if !defined?(RUBY_ENGINE) || RUBY_ENGINE != 'jruby'
+        begin
+          Signal.trap('USR1') do
+            puts 'Caught USR1; dumping a stack trace'
+            puts caller.map { |i| "  #{i}" }.join("\n")
+          end
+        rescue ArgumentError
```
mikz: Rubocop should scream about this. Why this error can happen? At least a comment explaining why it would be useful.

lib/3scale_toolbox/cli/error_handler.rb
eguzki: Some information like error message and stacktrace is printed to STDERR.

For providing support lots of information bits are dumped to a file. Have seen it in other projects and I think it is a good idea.

If working with ephemeral containers, this feature will be disabled, obviously.

eguzki:
moved error handling to `ThreeScaleToolbox::CLI` module

lib/3scale_toolbox/cli.rb
```diff
@@ -12,9 +14,139 @@ def self.load_builtin_commands
...
+
+  # Catches errors and prints nice diagnostic messages
+  class ErrorHandler
+    def self.error_watchdog(exit_on_error: true)
+      new.error_watchdog(exit_on_error: exit_on_error) { yield }
+    end
+
+    def error_watchdog(exit_on_error:)
+      # Set exit handler
+      %w[INT TERM].each do |signal|
```
mikz: This is still in the `ErrorHandler` module. It is something that is global and being executed in a block.

By moving it into the main module I did not simply mean move the `ErrorHandler` module contents to a different file, but move this code into the other module so it is obvious it is global.

eguzki: done

lib/3scale_toolbox/cli.rb
```diff
@@ -12,9 +14,137 @@ def self.load_builtin_commands
...
+  end
+
+  def self.error_watchdog(exit_on_error: true)
+    new.error_watchdog(exit_on_error: exit_on_error) { yield }
+  end
+
+  # Catches errors and prints nice diagnostic messages
+  def error_watchdog(exit_on_error:)
+    # Set exit handler
+    %w[INT TERM].each do |signal|
```
mikz: This is just moving this from one module from another. It does not address the fact that this is global action that is for example not reverted back.

IMO this class does not have one reason to change. It does error handling but also installs signal handlers for an interruption. Those are two different concepts that should be separate.

`error_watchdog` should basically just be `yield` + `rescue`. Installing signal handlers should be done in `self.run` before entering the `error_watchdog`.

lib/3scale_toolbox/cli.rb
```diff
@@ -12,9 +14,137 @@ def self.load_builtin_commands
...
+  private
+
+  def handle_error(error, exit_on_error:)
+    if expected_error?(error)
+      warn
+      warn "\e[1m\e[31mError: #{error.message}\e[0m"
+    else
+      print_error(error)
+    end
+    exit(1) if exit_on_error
```
mikz: This is passing a parameter 3 levels deep just to call exit. IMO this should be done in `self.run`. That is what I meant by doing it in the CLI module. Not moving the whole method, but just this action.

lib/3scale_toolbox/cli/error_handler.rb
```diff
@@ -0,0 +1,122 @@
...
+module ThreeScaleToolbox
+  module CLI
+    class ErrorHandler
+      def self.error_watchdog
+        new.error_watchdog { yield }
+      end
+
+      # Catches errors and prints nice diagnostic messages
+      def error_watchdog
+        error = false
```
mikz: Why not?

```ruby
def error_watchdog
  yield
  false
rescue StandardError, ScriptError => err
  handle_error err
  return err
end
```

exe/3scale
```diff
@@ -4,4 +4,4 @@ require '3scale_toolbox'
 
 args = ARGV.clone
 
-ThreeScaleToolbox::CLI.run args
+exit(ThreeScaleToolbox::CLI.run(args) ? 1 : 0)
```
mikz: Seems a bit odd to signal error with truthy return value.

Issue #38, opened 2018-06-19
Command to import service using OpenAPI / Swagger spec or RAML format


eguzki:
Using an API definition format like OpenAPI / Swagger or RAML you can automate some setups and save a lot of time in configuration.

When you import an API defintion, the following actions will be performed in the background:

1. Create a new service (unless you specify one)
1. Create methods in the 'Definition' section
1. Attach newly created methods to the 'Hits' metric
1. Create mapping rules and show them under API > Integration

andrewdavidmackenzie:
I'd scope the issue to focus on Open API Spec.
RAML, if desired, could be done as a separate effort.

PR #76, opened 2018-11-28
Import OpenAPI v2.0 command

eguzki:
import into your 3scale API OpenAPI 2.0 defiinition.

Try to create new service with OpenAPI definition. If `system_name` already used, update service.

`system_name` taken from `--target_system_name`. Default to OpenAPI `info.title` attribute.

- [x] import openapi commad
- [x] unittests 
- [x] integration tests
- [x] Documentation in README.md
- [x] Optional parameter `--target_system_name` for specifying `system_name`
- [x] OpenAPI spec can be a file or url
- [x] OpenAPI spec can be read from stdin
- [x] handle OpenAPI 2.0 parsing/validation errors
- [x] command is idempotent


eguzki:
Fixes https://github.com/3scale/3scale_toolbox/issues/38

mikz:
I think this might be the point to start thinking about a dry-run mode. Running potentially huge import with no option to roll back could have a devastating impact on someone's portal.

eguzki:
A couple of questions: 
- Dry-run mode for this command or for any command doing not read-only operations on remotes?
- For the specific case of importing OpenAPI spec case, what strategy do you propose for the implementation? On dry run, mocking 3scale client?

eguzki:
Finished all remaining tasks, PR ready to be reviewed @mikz

mikz:
@eguzki I think dry-run just for this command would be enough. Of course, it would be cool to have it in the others too, but not necessary I'd say.

Regarding mocking the client: I'm not sure if you can mock the client (or you can stub just POST/PUT/DELETE operations). I think the dry mode could be "read only mode". Just read what is the current state and print out the plan (... to be deleted, ... to be changed from .. to ...).

docs/openapi.md
```diff
@@ -0,0 +1,82 @@
+## Import API definition to 3scale from OpenAPI definition
+
+Features:
+
+* OpenAPI __2.0__ specification (f.k.a. __swagger__)
+* Create a new service. New service name will be taken from openapi definition `info.title` field.
+* Update existing service, providing `SERVICE_ID` with `--service` option.
+* Create methods in the 'Definition' section. Method names are taken from `operation.operationId` field.
+* Attach newly created methods to the *Hits* metric.
+* Create mapping rules and show them under `API > Integration`.
```
mikz: Is it going to keep the `/ => hits` mapping?

exe/3scale
```diff
@@ -1,6 +1,14 @@
...
-require '3scale_toolbox'
+def suppress_warnings
+  original_verbosity = $VERBOSE
+  $VERBOSE = nil
+  result = yield
+  $VERBOSE = original_verbosity
+  result
+end
+
+suppress_warnings { require '3scale_toolbox' }
```
mikz: What are the warnings?

exe/3scale
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env ruby
 
-require '3scale_toolbox'
+def suppress_warnings
+  original_verbosity = $VERBOSE
+  $VERBOSE = nil
+  result = yield
+  $VERBOSE = original_verbosity
```
mikz: I think this should rather use `ensure` block to reset back the verbosity.

lib/3scale_toolbox/commands/import_command/openapi/command.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          def self.command
+            Cri::Command.define do
+              name        'openapi'
+              usage       'openapi [opts] -d <dst> <spec>'
+              summary     'Import API defintion in OpenAPI specification'
+              description 'Using an API definition format like OpenAPI, import to your 3Scale API'
+
+              option  :d, :destination, '3scale target instance. Format: "http[s]://<authentication>@3scale_domain"', argument: :required
+              option  :s, :service, '<service_id> of your 3Scale account', argument: :required
```
mikz: 👀  3Scale

lib/3scale_toolbox/commands/import_command/openapi/command.rb
```diff
@@ -0,0 +1,66 @@
...
+
+          def run
+            context = create_context
+            if options[:service]
+              context[:service] = Entities::Service.new(id: options[:service],
+                                                        remote: context[:threescale_client])
+            end
+
+            tasks = []
+            tasks << CreateServiceStep.new(context) unless options[:service]
```
mikz: So, it is not possible to create and update the service by the same command?

Then eventually the pipeline importing the spec would have to know if the service exists or not and that can be hard.

Have you considered passing service system_name instead of the id, so it can be used to create/update the same service?

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
```diff
@@ -0,0 +1,35 @@
...
+              'metric_id' => metric_id
+            }
+          end
+
+          def http_method 
+            operation[:verb].upcase
+          end
+
+          def pattern
+            operation[:path]
```
mikz: Is the pattern compatible with what mapping rules support?

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
```diff
@@ -0,0 +1,50 @@
...
+          def read_file(resource)
+            [File.read(resource), { format: File.extname(resource) }]
+          end
+
+          def read_stdin(_resource)
+            content = STDIN.read
+            # will try parse json, otherwise yaml
+            format = :json
+            begin
+              JSON.parse(content)
```
mikz: JSON is a subset of YAML, so parsing it as YAML should work.

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
```diff
@@ -0,0 +1,50 @@
...
+            end.call(resource)
+          end
+
+          # Detect format from file extension
+          def read_file(resource)
+            [File.read(resource), { format: File.extname(resource) }]
+          end
+
+          def read_stdin(_resource)
+            content = STDIN.read
```
mikz: I think you can collapse all these implementations into `YAML.parse(ARGF)`

spec/shared_contexts.rb
```diff
@@ -48,3 +48,109 @@ def get_plugin_content(command_class_name, command_name)
...
+end
+
+RSpec.shared_context :import_oas_real_cleanup do
+  after :each do
+    service.delete_service
+  end
+end
+
+RSpec.shared_context :import_oas_stubbed_api3scale_client do
+  puts '============ RUNNING STUBBED 3SCALE API CLIENT =========='
```
mikz: This is not printed when running the client but when importing the module. It would have to be within `before(:all)` block.

spec/shared_examples.rb
```diff
@@ -0,0 +1,54 @@
...
+    ).to be_empty
+  end
+
+  it 'mapping rules are created' do
+    expect { subject }.to output.to_stdout
+    expect(expected_mapping_rules.size).to be > 0
+    # test Set(service.mapping_rules) includes Set(expected_mapping_rules)
+    # with a custom identity method for mapping_rules
+    expect(
+      ThreeScaleToolbox::Helper.array_difference(expected_mapping_rules, service.mapping_rules) do |el1, el2|
```
mikz: I think it would make sense to extract this into a matcher.

```ruby
expect(expected_mapping_rules).to have_no_difference(service.mapping_rules).comparing_keys(mapping_rule_keys)
```

docs/openapi.md
eguzki: Yeah, target service mapping rules are not deleted. Both when new service is created or existing service is updated.

It is behaving with semantics of HTTP `POST`. Should be HTTP `PUT` semantics?

exe/3scale
eguzki: For ruby 2.5, deprecation warnings from `colored` and `hashie` (dependency of swagger lib) libraries

```
3scale_toolbox/vendor/bundle/ruby/2.5.0/gems/colored-1.2/lib/colored.rb:45: warning: shadowing outer local variable - value
3scale_toolbox/vendor/bundle/ruby/2.5.0/gems/hashie-3.3.2/lib/hashie/extensions/coercion.rb:16: warning: constant ::Fixnum is deprecated
3scale_toolbox/vendor/bundle/ruby/2.5.0/gems/hashie-3.3.2/lib/hashie/extensions/coercion.rb:16: warning: constant ::Bignum is deprecated
3scale_toolbox/vendor/bundle/ruby/2.5.0/gems/hashie-3.3.2/lib/hashie/extensions/coercion.rb:17: warning: constant ::Fixnum is deprecated
3scale_toolbox/vendor/bundle/ruby/2.5.0/gems/hashie-3.3.2/lib/hashie/extensions/coercion.rb:17: warning: constant ::Bignum is deprecated
```

warnings are available running tests, not to forget about them.

docs/openapi.md
mikz: I'm not sure if this is about PUT or POST. 

Having the default mapping rule `GET / => hits` will match every GET request and also the mapping rule for the operation ID and reporting both values to backend resulting in double reporting.

exe/3scale
mikz: Then maybe just wrap loading colored and hashie in supress_warnings ?

lib/3scale_toolbox/commands/import_command/openapi/command.rb
eguzki: Since the are no clear specs on the feature, this is another decision taken inspired from
https://github.com/3scale/3scale-cli. It is exactly the same behavior.

Even more, on the current implementation, since `system_name` is taken from swagger spec, trying to import twice the swagger definition on the same tenant (creating new services) will fail on the second try because `system_name` will be taken. It makes sense for me, so I implemented like that.

IMO the user should know whether new service is being created or a service is being updated (hence `service_id` needs to be known in advance). But if it has some value to abstract that decision from the user, providing `system_name` would indeed be the way to go.

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
eguzki: haven't checked that. Is it documented what mapping rules support in `pattern` field? Send link please.

lib/3scale_toolbox/commands/import_command/openapi/command.rb
mikz: I think when used in pipeline (that is something we want to support) there is no way to know if you are creating or updating a service.
We would have to provde a command to get a service and then let customers write the logic.

IMO we should rather offer way for create or update out of the box to help automation.

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
mikz: The documentation: https://github.com/3scale/porta/blob/8022f744c58a8a64f6fcfb8d2f2129fd9eaad74d/config/locales/en.yml#L530-L557

The implementation: https://github.com/3scale/apicast/blob/c3e431df20f65686798d1e67d8c4e9f568ef9c08/gateway/src/apicast/mapping_rule.lua#L30-L35

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
eguzki: I would not say JSON is a subset of YAML :)

Anyway, it is true, indeed, that yaml parser in ruby parses json syntax.

The swagger parser interface requires two params, content stream and format (`json` or `yaml`). `read_stdin` method has no way to find out format other than checking if json parser fails. The result is not even used. So we need to parse to find out format.

Or we can always tell swagger parser it is yaml, even when it is json. But I guess it is ugly and haven't checked if it works. But the interface makes clear distinction between `json` nad `yaml`

mikz: YAML is a superset of JSON. Not only in Ruby. Everywhere.

Yes, you can use YAML just fine https://github.com/swagger-rb/swagger-rb/blob/51b0a68e5c24e9b8eeefb3467700da58d5cac884/lib/swagger/parsers.rb#L26.

You can remove almost everyting, use ARGF to read all files and stdin, always use YAML and just detect if there is an URL you need.

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
eguzki: First, as mentioned in https://github.com/3scale/3scale_toolbox/pull/76#discussion_r241045479, we still need to find out format.

Second, ARGF is like `cat`. Nice but what if we would like to add some  extra positional param?  Less flexible as far as I see.

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
mikz: And you can just pass already parsed document to the swagger library: https://github.com/swagger-rb/swagger-rb/blob/51b0a68e5c24e9b8eeefb3467700da58d5cac884/lib/swagger.rb#L25

That makes more sense anyway to separate concerns of parsing/loading and the actual spec.

spec/shared_examples.rb
eguzki: Nice
I would say:

```
expect(expected_mapping_rules).to be_subset_of(service.mapping_rules).comparing_keys(mapping_rule_keys)
```

eguzki:
Thanks for the good feedback on relevant issues! The show must go on

docs/openapi.md
eguzki: Will remove existing mapping_rules as default behavior. Maybe with a flag to avoid removing when set.

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
mikz: Format is not an issue as already mentioned there.

What extra positional arguments? Do you have an example?

It really depends what CRI does to the ARGV and if leaves unprocessed arguments.
That is not as big a deal anyway.

spec/shared_examples.rb
mikz: Yep, that looks good!

exe/3scale
eguzki: But it is interesting to uncover any happening warning, at least when testing. That's why it was implemented in `exe/3scale` entrypoint script

mikz: Thats a good point. But once we can't do anything about the warning, we should supress it no? Even in tests it would be just distraction without any real value.

lib/3scale_toolbox/commands/import_command/openapi/command.rb
eguzki: Good, then `--system-name` could be optional parameter. It not provided, `system_name` could be set from `info.title` from openAPI definiton.

Having the `system_name`, the toolbox will check whether service exists and run create/update accordingly. 

In fact, we could do the same with `update service` and `copy service`. The toolbox will, out of the box, perform required tasks from provided `system_name`. The two commands lead to confusion and they do the same task, one on newly created service and the other on existing one.

mikz: So, even without supplying `--system-name` it would take it from the `info.title` and update the existing service, right?

eguzki: if the service would exist, then update. If not, create one. This behavior could be independent from the source of `system_name`

mikz: 👍  sounds perfect!

exe/3scale
eguzki: IMO warnings in development environments are always nice. If you suppress unconditionally and tomorrow we upgrade library, potentially interesting warnings could be shadowed.

mikz: People usually freak out when they see warnings and give them unreasonable importance.
This is happening to us in APIcast. People don't see the real error for some useless warnings in the output.

eguzki: I would expect that behavior from production environment users. Not from developers on development environments.

mikz: And yet, people who try to contribute often get stuck on these.

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
eguzki: Mapping rules accepted expression set includes openapi 2.0 path expressions.

The field name MUST begin with a slash and Path Templating is allowed.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-fields

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
eguzki: Using `YAML.load` and removing format tracking login
 https://github.com/3scale/3scale_toolbox/pull/76/commits/8bba8577d2b1b76dfa3a823537dab9abe8bddd8a

lib/3scale_toolbox/commands/import_command/openapi/resource_reader.rb
eguzki: according to https://robots.thoughtbot.com/rubys-argf
```
ARGF will interpret all elements of the ARGV array as filenames and when read will produce a concatenation of the contents of these files. If ARGV is empty, then ARGF reads from standard input.
```
Commands and subcommands will be treated as filenames, so hard to integrate with CRI command framework. `ARGV` needs to be processed first. Moreover, options and flags need to be removed from `ARGV` before calling `ARGF.read`.

WRT extra positional argument, imagine we want to pass 3scale instance remote name as positional argument instead of option with `-d`. It would be treated like filename or it would need to be removed before calling `ARGF.read`.

IMO unnecessary complication for the provided little value.

eguzki:
* Verified the patterns in the spec match what mapping rules can match
* Update service from `system_name`. If it does not exist, create it first. `system_name` can be passed as option parameter and defaults to `info.title` from openapi spec.
* All existing mapping rules are deleted before importing new API definition. Methods left untouched if exist.

eguzki:
rebased to current master, since it diverged too much from branch base

eguzki:
ping @mikz

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
mikz: I'm not sure it works. 

Defining a path `/pets` would match `/pets/{id}` too. Is that desired? Our expressions are not the exact match to the whole string, just matching the beginning.

spec/unit/copy_service_spec.rb
```diff
@@ -22,6 +22,7 @@
       # Entities::Service instance stub
       service = double('Entities::Service instance')
```
mikz: Would using [object double](https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/verifying-doubles/using-an-object-double) simplify the test, do you don't have to setup `id` for example?

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
eguzki: Cannot find path matching spec in OAS 2.0 spec. I just found [clarifications on path overlapping issues](https://github.com/OAI/OpenAPI-Specification/issues/1127)

From common sense, I would expect OAS 2.0 spec to do exact match. Just a guess.

Since apicast does just prefix matching, I guess we need to specify somehow end of string when creating mapping rule in 3scale. That only if my assumption about exact match on OAS 2.0 is correct. Is there a way to do that? Adding '/' in the end would not work.

mikz: @eguzki it is described in the documentation mentioned in https://github.com/3scale/3scale_toolbox/pull/76#discussion_r241044806 (using $ at the end)

spec/unit/copy_service_spec.rb
eguzki: I do not think it would simplify the test. I would have to construct the service object and pass `id` and `remote` objects (doubles?) to the constructor. On the other hand it would check that any stubbed methods on the double also exist on the template, which it is not done now. 

I consider this is not a big deal for this test.

mikz: Object double verifies that you can't stub a method that does not exist. I think [instance_double](https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/verifying-doubles/using-an-instance-double) does the same and does not require the constructor.

eguzki:
regarding dry-run mode, I opened issue on that https://github.com/3scale/3scale_toolbox/issues/87 to leave out of the scope of this PR

eguzki:
- [x] applied strict matching on mapping rule patterns
- [x] use `instance_double` when appropriate to check sutbbed methods

PR #77, opened 2018-11-30
Using remote names from commands

eguzki:
- [x] Using remote names from commands
- [x] Doc remote management commands (list, add, remove, rename)
- [x] Doc using remote names from commands


eguzki:
@andrewdavidmackenzie doc mistakes review on you :)

README.md
```diff
@@ -77,13 +81,15 @@ OPTIONS FOR COPY
...
 ```shell
-3scale copy service NUMBER --source=https://provider_key@foo-admin.3scale.net --destination=https://provider_key@foo2-admin.3scale.net
+3scale copy service NUMBER --source=foo --destination=https://access_token@foo2-admin.3scale.net
 ```
 
 ### Update a service
 
 Will update existing service, update proxy settings, metrics, methods, application plans and mapping rules.
 
+3Scale instances can be either a [URL](docs/remotes.md#remote-urls) or the name of a [remote](docs/remotes.md).
```
mikz: 👀

docs/remotes.md
```diff
@@ -0,0 +1,82 @@
...
+
+The following syntax is used:
+
+```
+http[s]://<provider_key>|<access_token>@<3scale-instance-domain>
+```
+
+## Commands
+
+### List
```
mikz: It might be better to actually have a subcommand `list` and `3scale remote` just show a list of subcommands.

lib/3scale_toolbox/helper.rb
```diff
@@ -1,24 +1,36 @@
 module ThreeScaleToolbox
-  module Helper
-    def self.compare_hashes(first, second, keys)
-      keys.map { |key| first.fetch(key, nil) } == keys.map { |key| second.fetch(key, nil) }
-    end
+  class Helper
```
mikz: Why is it a class? Do you plan make instance of it?

lib/3scale_toolbox/remotes.rb
```diff
@@ -1,81 +1,91 @@
...
-    def valid_remote?(remote)
-      remote.is_a?(Hash) \
-        && remote.key?(:endpoint) \
-        && remote.key?(:auth_key)
+    def valid?(remote)
+      remote.is_a?(Hash) && remote.key?(:endpoint) && remote.key?(:auth_key)
     end
 
-    def validate_remotes?(remotes)
+    def validate?(remotes)
```
mikz: Calling the method `validate?` suggests the validation can be on / off. 

Drop the `?`.

spec/unit/remotes_spec.rb
```diff
@@ -1,39 +1,159 @@
...
+  RSpec.shared_examples 'raises invalid url error' do
+    it 'raises error' do
+      expect { subject.add_uri('some_name', origin) }.to raise_error(ThreeScaleToolbox::InvalidUrlError,
+                                                                     /invalid url/)
+    end
   end
 
-  subject { test_class.new }
+  let(:yaml_content) do
+    <<~HEREDOC
```
mikz: if you use `YAML` instead of `HEREDOC` then some editors will be able to actually do syntax highlighting on the string.

spec/unit/remotes_spec.rb
```diff
@@ -1,39 +1,159 @@
...
+      ---
+        :remotes:
+          remote_1:
+            :endpoint: https://1.example.com
+            :auth_key: '123456789'
+    HEREDOC
+  end
+  let(:config_file) do
+    File.join(tmp_dir, '.3scalerc').tap do |config_file|
```
mikz: This would be way nicer with Pathname.


```ruby
tmp_dir.join('.3scalerc').write(yaml_content)
```

spec/unit/remotes_spec.rb
```diff
@@ -1,39 +1,159 @@
...
+      File.open config_file, 'w' do |file|
+        file.write(yaml_content)
+      end
+    end
+  end
+  let(:config) { ThreeScaleToolbox::Configuration.new(config_file) }
+  let(:endpoint) { 'https://1.example.com' }
+  let(:authkey) { '123456789' }
```
mikz: There is no concept of `authkey` in 3scale.

* access_token
* provider_key

Or more general terms for Basic auth:

* user
* password
* userinfo

https://tools.ietf.org/html/rfc3986#section-3.2.1

docs/remotes.md
eguzki: `git remote` gives the list of remotes

Then you have `git remote add`, `git remote rename`, etc...

I think it kind of makes sense.

for list of subcommands, just 
```shell
$ 3scale remote -h
NAME
    remote - 3scale CLI remotes

USAGE
    3scale remote <command> [options]

DESCRIPTION
    3scale CLI command to manage your remotes

SUBCOMMANDS
    add        3scale CLI remote add
    remove     3scale CLI remote remove
    rename     3scale CLI remote rename

OPTIONS FOR 3SCALE
    -c --config-file=<value>      3scale CLI configuration file (default:
                                  /home/eguzki/.3scalerc.yaml)
    -h --help                     show help for this command
    -k --insecure                 Proceed and operate even for server
                                  connections otherwise considered insecure
    -v --version                  Prints the version of this command
```

lib/3scale_toolbox/helper.rb
eguzki: No, but I did not plan to include it either. 

Changing to `module`

spec/unit/remotes_spec.rb
eguzki: I got the name from https://github.com/3scale/3scale_toolbox/pull/53#discussion_r230045161

That is a key given to 3scale client. According to 3scale client constructor:
https://github.com/3scale/3scale-api-ruby/blob/585c09f69ab57e62a48d17d05e9132c275719f6e/lib/3scale/api.rb#L8

Some `provider_key` has to be given. Since `access_token`s are also valid, IMO it seems `token` or `auth_key` to be more appropriate. In the request whether it is user information or auth header or some other kind of token is implementation detail. 

What do you propose? `user` or `password` does not seem to fit very well. `userinfo`??

docs/remotes.md
mikz: Git should not be called out as a well designed CLI.
I know what it does and I'm saying it might be better if the `3scale remote`  to print the list of subcommands available and use `3scale remote list` to list them, because of discoverability.

I'd say `kubectl` and `oc` are well designed CLI that give you at each step enough information.

eguzki: let me do this change in another PR

spec/unit/remotes_spec.rb
eguzki: Sorry, do not see the difference between `authkey` and `authentication`. Moreover, IMO `authkey` is more meaninful. `authentication` is more about a method or process for identifying a user, while `authkey` is about id item of that process. 

About changing the format remote credentials are stored, IMO this is out of scope of this PR. Happy to discuss about formats. Having specific field for the token or having as part of the userinfo in the url.

mikz: The problem with `authkey` is that is is made up word and similar to `app_key` that is already used by 3scale.

eguzki: changed to `authentication` :+1:

lib/3scale_toolbox/helper.rb
```diff
@@ -1,24 +1,36 @@
 module ThreeScaleToolbox
   module Helper
-    def self.compare_hashes(first, second, keys)
-      keys.map { |key| first.fetch(key, nil) } == keys.map { |key| second.fetch(key, nil) }
-    end
+    class << self
```
mikz: Usually just `module_function` call on top of the module solves this.

PR #85, opened 2018-12-24
Some wording changes and fix error in command list

andrewdavidmackenzie:
I made some small wording change suggestions.
We should come up with a new term for "instance" I think, as generally people think that an instance is a separate install.
Added "list" to the synopsis


eguzki:
May conflict with https://github.com/3scale/3scale_toolbox/pull/84.

Let's merge this first and I'll rebase #84

PR #88, opened 2019-01-31
docs/openapi: fix doc to update service by system_name

eguzki:
Documentation update to be consistent with last implementation updates


eguzki:
ping @mkudlej

PR #97, opened 2019-03-03
Feature/copy update activedocs

eguzki:
`3scale copy service` and `3scale update service` commands add activedocs to the list of entities to copy/update.

Needs new release of `3scale-api-ruby` with the following PR merged
https://github.com/3scale/3scale-api-ruby/pull/46


eguzki:
ready for review @mikz @miguelsorianod

lib/3scale_toolbox/tasks/copy_activedocs_task.rb
```diff
@@ -0,0 +1,24 @@
...
+          source_activedoc.clone.tap do |target_activedoc|
+            target_activedoc.delete('id')
+            target_activedoc['service_id'] = target.id
+            target_activedoc['system_name'] = "#{target_activedoc['system_name']}#{target.id}"
+          end
+        end
+
+        target_activedocs.each do |ad|
+          res = target.remote.create_activedocs(ad)
+          raise Error, "ActiveDocs has not been created. Errors: #{res['errors']}" unless res['errors'].nil?
```
mikz: It is hard to know which class the Error resolves to and would be nicer to use the full namespace.

eguzki: updated to use full namespace

PR #100, opened 2019-03-05
Fix/target system name optional

### LINKED ISSUE #99: copy service  "-t" argument is confusing
Documentation for "copy service" makes it appear as though the "-t" or "--target-system-name" argument is optional, but it appears to be required.  The purpose of this argument is also not very clear from the documentation.

It would be best if this argument was made optional and the tool re-used the system name from the source service by default.  The argument could then be used to override that behavior.

At the very least, please update the documentation to reflect the current behavior of the tool.

eguzki:
In `3scale copy service` command, ''--t -target_system_name` optional param. Defaults to source service system name.

Fixes #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,8 @@ OPTIONS
                                          remote name
     -s --source=<value>                  3scale source instance. Url or
                                          remote name
-    -t --target_system_name=<value>      Target system name
+    -t --target_system_name=<value>      Target system name. Default to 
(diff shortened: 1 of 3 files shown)


eguzki:
refactored a little bit the test for clearness.

ready for review @mikz

eguzki:
added tests  to verify it is indeed optional and defaults to the source system name.

PR #102, opened 2019-03-06
import openapi: fix when missing operation id in spec

eguzki:
in OAS 2.0, when there is no `operation_id` attribute, i.e.

```json
....
"paths": {
    "/products": {
      "get": {
        "summary": "Product Types",
        "description": "The Products endpoint returns information about the Uber products offered at a given location. The response includes the display name and other details about each product, and lists the products in the proper display order.",
        "parameters": [
          {
            "name": "latitude",
            "in": "query",
            "description": "Latitude component of location.",
            "required": true,
            "type": "number",
            "format": "double"
          },
          {
            "name": "longitude",
            "in": "query",
            "description": "Longitude component of location.",
            "required": true,
            "type": "number",
            "format": "double"
          }
        ],
        "tags": [
          "Products"
        ],
        "responses": {
          "200": {
            "description": "An array of products",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Product"
              }
            }
          },
          "default": {
            "description": "Unexpected error",
            "schema": {
              "$ref": "#/definitions/Error"
            }
          }
        }
      }
    }
....
```

method names are computed:

```ruby
"#{operation[:verb]}#{operation[:path].tr('/{}', '')}"
```

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi/method.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/method.rb
@@ -12,11 +12,15 @@ module ThreeScaleToolbox
           end
 
           def friendly_name
-            operation[:operationId]
+            operation[:operationId] || operation_id
           end
 
           def system_name
             friendly_name.downcase
+          end
+
+          def operation_id
+            "#{operation[:verb]}#{operation[:path].gsub(/[^\w]/, '')}"
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/3scale_toolbox/commands/import_command/openapi/method.rb
```diff
@@ -12,12 +12,16 @@ def method
...
-            operation[:operationId]
+            operation[:operationId] || operation_id
           end
 
           def system_name
             friendly_name.downcase
           end
+
+          def operation_id
+            "#{operation[:verb]}#{operation[:path].tr('/{}', '')}"
```
mikz: Woudn't it be safer to strip out everything other than `/\w/` ? `.gsub(/[^\w]/, '')`

eguzki: done

Issue #90, opened 2019-02-05
RFE: Create an ActiveDocs with updated environmental fields when importing a service from OpenAPI


nmasse-itix:
# Current Situation

When importing a service from an OpenAPI Specification, using:

```sh
3scale import openapi -t echo_api -d https://$ACCESS_TOKEN@$TENANT-admin.3scale.net examples/echo-api.yaml
```

The service is created but no ActiveDocs is created. 

# Proposed Changed

The OpenAPI Specification used to create the service is pushed as an ActiveDocs file, using the same system_name as the service. 

This OpenAPI Specification is patched to take into account the new environment: 
- the `host` field is updated with the hostname component of the public production base URL
- the `schemes` field is updated with the scheme component of the public production base URL
- the `basePath` field is updated if rewriting rules have been set
- the `securityDefinitions/*/authorizationUrl` and `securityDefinitions/*/tokenUrl` fields are updated with the token and authorize endpoints listed in the OpenID Connect wellknown endpoint (when OIDC is in use)

# Business need

Currently, the missing ActiveDocs implies for the customer to have a tedious manual step to update the OpenAPI Spec and upload it. This is needed to achieve a "ready-to-use" API on the Developer Portal.

This is something I'm currently doing in the [Ansible playbooks](https://github.com/nmasse-itix/threescale-cicd/blob/f22a0fc3cb70f91cc6edc11117bb14f3940b5730/templates/rewritten_openapi.j2).

sgutierr:
Older CLI tool offered this feature

PR #103, opened 2019-03-06
Import oas include activedocs

eguzki:
The OpenAPI Specification used to create the service is pushed as an ActiveDocs file, using the same system_name as the service.

This PR partially covers issue #90 
Still missing: This OpenAPI Specification is patched to take into account the new environment:
* the host field is updated with the hostname component of the public production base URL
* the schemes field is updated with the scheme component of the public production base URL
* the basePath field is updated if rewriting rules have been set
* the securityDefinitions/*/authorizationUrl and securityDefinitions/*/tokenUrl fields are updated with the token and authorize endpoints listed in the OpenID Connect wellknown endpoint (when OIDC is in use)

The missing parts will be included in another PR

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
@@ -1,0 +1,52 @@
+module ThreeScaleToolbox
+  module Commands
+    module ImportCommand
+      module OpenAPI
+        class CreateActiveDocsStep
+          include Step
+
+          def call
+            active_doc = {
+              name: api_spec.title,
+              system_name: activedocs_system_name,
+              service_id: service.id,
+              body: resource.to_json,
+              description: api_spec.description
+            }
+
+            res = threescale_client.create_activedocs(active_doc)
+            # Make operation indempotent
+            if (errors = res['errors'])
+              raise ThreeScaleToolbox::Error, "ActiveDocs has not been created. #{errors}" \
+                unless system_name_already_taken_error? errors
+
+              # if activedocs system_name exists, ignore error, update activedocs
+              puts 'Activedocs exists, update!'
+              update_res = threescale_client.update_activedocs(find_activedocs_id, active_doc)
+              raise ThreeScaleToolbox::Error, "ActiveDocs has not been updated. #{update_res['errors']}" unless update_res['errors'].nil?
+            end
+          end
+
+          private
+
+          def activedocs_system_name
+            @activedocs_system_name ||= service.show_service['system_name']
+          end
+
+          def find_activedocs_id
+            activedocs = get_current_service_activedocs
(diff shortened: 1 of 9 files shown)


lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          def activedocs_system_name
+            @activedocs_system_name ||= service.show_service['system_name']
+          end
+
+          def find_activedocs_id
+            activedocs = get_current_service_activedocs
+            raise ThreeScaleToolbox::Error, "Could not find activedocs with system_name: #{activedocs_system_name}" if activedocs.empty?
+
+            activedocs[0]['id']
```
mikz: Might be better to use `.dig` ?

lib/3scale_toolbox/commands/import_command/openapi/step.rb
```diff
@@ -38,6 +38,18 @@ def target_system_name
             # could be nil
             context[:target_system_name]
           end
+
+          def resource
+            context[:api_spec_resource]
+          end
+
+          def system_name_already_taken_error?(error)
+            return false if error.nil?
```
mikz: `error && Array(error['system_name']).any? { |msg| msg.match /has already been taken/ }`

lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
eguzki: updated

lib/3scale_toolbox/commands/import_command/openapi/step.rb
eguzki: cool! updated!

lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
```diff
@@ -0,0 +1,52 @@
...
+              name: api_spec.title,
+              system_name: activedocs_system_name,
+              service_id: service.id,
+              body: resource.to_json,
+              description: api_spec.description
+            }
+
+            res = threescale_client.create_activedocs(active_doc)
+            # Make operation indempotent
+            unless res['errors'].nil?
```
mikz: unless nil, so if errors? 

`if (errors = res['errors'])` and you can use just `errors` later in the branch.

lib/3scale_toolbox/commands/import_command/openapi/create_method_step.rb
```diff
@@ -13,12 +13,10 @@ def call
               # if method system_name exists, ignore error and get metric_id
               # Make operation indempotent
               unless res['errors'].nil?
```
mikz: Same comment as 
https://github.com/3scale/3scale_toolbox/pull/103/files#r263431926

eguzki:
changed those unless + nil conditions

PR #109, opened 2019-03-15
ActiveDocs created in published state

### LINKED ISSUE #108: ActiveDocs created in "hidden" state and there is no switch to make it visible
# Current behavior

When I'm importing an OpenAPI Spec using the `master` branch:
```raw
$ bundle exec 3scale import openapi examples/echo-api.yaml -d nmasse-redhat -t echo_api
Updated service id: 2555417776440, name: Echo API
destroying all mapping rules
Created GET /$ endpoint
```

The ActiveDocs is created in "hidden" state and there is no switch in the `3scale import openapi` to make it visible:
Screen Shot 2019-03-14 at 16 57 10

# Expected behavior

The `3scale import openapi` command creates/updates the ActiveDocs and it is visible by default.

eguzki:
When importing OAS, ActiveDoc is created by default on `visible` state

Fixes #108

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi.rb
@@ -27,6 +27,7 @@ module ThreeScaleToolbox
 
               option  :d, :destination, '3scale target instance. Format: "http[s]://<authentication>@3scale_domain"', argument: :required
               option  :t, 'target_system_name', 'Target system name', argument: :required
+              flag    nil, 'activedocs-hidden', 'Create ActiveDocs in hidden state'
               param   :openapi_resource
 
               runner OpenAPISubcommand
@@ -34,9 +35,6 @@ module ThreeScaleToolbox
           end
 
           def run
-            openapi_resource = load_resource(arguments[:openapi_resource])
-            context = create_context(openapi_resource)
-
             tasks = []
             tasks << CreateServiceStep.new(context)
             tasks << CreateMethodsStep.new(context)
@@ -50,12 +48,18 @@ module ThreeScaleToolbox
 
           private
 
-          def create_context(openapi_resource)
+          def context
+            @context ||= create_context
+          end
+
+          def create_context
+            openapi_resource = load_resource(arguments[:openapi_resource])
             {
               api_spec_resource: openapi_resource,
               api_spec: ThreeScaleApiSpec.new(load_openapi(openapi_resource)),
               threescale_client: threescale_client(fetch_required_option(:destination)),
-              target_system_name: options[:target_system_name]
+              target_system_name: options[:target_system_name],
+              activedocs_published: !options[:'activedocs-hidden']
             }
           end
 
--- a/lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
@@ -11,7 +11,8 @@ module ThreeScaleToolbox
               system_name: activedocs_system_name,
               service_id: service.id,
               body: resource.to_json,
-              description: api_spec.description
+              description: api_spec.description,
+              published: true,
             }
 
             res = threescale_client.create_activedocs(active_doc)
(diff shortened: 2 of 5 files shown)


lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
```diff
@@ -11,7 +11,8 @@ def call
               system_name: activedocs_system_name,
               service_id: service.id,
               body: resource.to_json,
-              description: api_spec.description
+              description: api_spec.description,
+              published: true,
```
mikz: Shouldn't this be configurable from the CLI?

eguzki: Is there a use case for creating `hidden`? I do not want to solve potential imaginary problems

mikz: Simply because I don't want my developers to see it, yet. 

It is the same use case why we have the feature at all.

eguzki: will add a flag

lib/3scale_toolbox/commands/import_command/openapi/create_activedocs_step.rb
```diff
@@ -11,7 +11,8 @@ def call
               system_name: activedocs_system_name,
               service_id: service.id,
               body: resource.to_json,
```
mikz: When touching this you might want to use JSON.pretty_generate, so the JSON looks ok to humans.
Because you are already decoding it, so the original formatting is gone.

eguzki: but this json serialization will not be consumed by humans, just 3scale api that requires json serialization.

mikz: It is, it appears in the UI when you edit the spec.

mikz: I think there are 3 options (ordered by their correctness in my eyes)

1. upload the original bytes and keep the original whitespace
2. upload a pretty printed copy, so humans can actually read that file in the UI 
3. upload a compacted copy for computers

eguzki: ok, let it be the original bytes then

eguzki: 3scale API only accepts JSON serialization format.

So I guess, best option is upload a pretty printed copy

eguzki: I mean, pretty json serialization

lib/3scale_toolbox/commands/import_command/openapi.rb
```diff
@@ -27,16 +27,14 @@ def self.command
 
               option  :d, :destination, '3scale target instance. Format: "http[s]://<authentication>@3scale_domain"', argument: :required
               option  :t, 'target_system_name', 'Target system name', argument: :required
+              flag    nil, 'activedocs-hidden', 'Create ActiveDocs in hidden state'
```
mikz: Isn't it weird that just above it is with underscores and this is with a dash?

eguzki:
```

Comment (eguzki) [reply lib/3scale_toolbox/commands/import_command/openapi.rb]: GNU style guide says dashes.
```
GNU adds long options to these conventions. Long options consist of ‘--’ followed by a name made of alphanumeric characters and dashes. Option names are typically one to three words long, with hyphens to separate words. Users can abbreviate the option names as long as the abbreviations are unique.
```

`target_system_name` is a mistake, but cannot be changed now

mikz: I see there are already two other instances of using dashes, so it is already a mixed bag. Ok.

PR #115, opened 2019-03-20
verbose mode on 3scale-api-ruby client calls

### LINKED ISSUE #67: verbose mode
It will be great to have verbose mode so user can watch command progress or/and operations/api calls done during command execution. I think this feature is essential for help user by another person in case of any troubles.

eguzki:
Verbose mode for debugging purposes.
Currently just PoC to be reviewed. Only activated for 3scale API calls, showing parameters sent and returned data, just at library level, not http level.

Method name, arguments and returned result are logged, together with call spent time . Truncated for readability.

Execution example:
```shell
3scale --verbose import openapi -d supertest https://petstore.swagger.io/v2/swagger.json
-- call create_service args |[{"name"=>"Swagger Petstore", "descriptio| response |{"errors"=>{"system_name"=>["has already | - (0.466925691s)
-- call list_services args |[]| response |[{"id"=>2555417723913, "name"=>"Default A| - (0.386342888s)
-- call update_service args |[2555417778023, {"name"=>"Swagger Petstor| response |{"id"=>2555417778023, "name"=>"Swagger Pe| - (1.341697979s)
Updated service id: 2555417778023, name: Swagger Petstore
-- call list_metrics args |[2555417778023]| response |[{"id"=>2555418192724, "name"=>"hits", "s| - (1.320869346s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (1.255788086s)
-- call list_metrics args |[2555417778023]| response |[{"id"=>2555418192724, "name"=>"hits", "s| - (0.421487273s)
-- call list_methods args |[2555417778023, 2555418192724]| response |[{"id"=>2555418192725, "name"=>"updatepet| - (0.527233717s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.41484545s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.853302026s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.3730865s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.891086951s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.818100019s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.397594189s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.38975787s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.983790828s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.548324586s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.945861828s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.999151823s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.559096439s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.550227549s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (1.152494071s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.384814473s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.485156714s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.54712834s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (0.387559833s)
-- call create_method args |[2555417778023, 2555418192724, {"friendly| response |{"errors"=>{"system_name"=>["has already | - (1.005019602s)
destroying all mapping rules
-- call list_mapping_rules args |[2555417778023]| response |[{"id"=>379740, "metric_id"=>255541819272| - (0.520080307s)
-- call delete_mapping_rule args |[2555417778023, 379740]| response |true| - (1.061655846s)
-- call delete_mapping_rule args |[2555417778023, 379741]| response |true| - (0.438651289s)
```

Fixes #67

### DIFF
--- a/lib/3scale_toolbox/proxy_logger.rb
+++ b/lib/3scale_toolbox/proxy_logger.rb
@@ -1,0 +1,20 @@
+module ThreeScaleToolbox
+  class ProxyLogger < BasicObject
+    def initialize(subject)
+      @subject = subject
+    end
+
+    def method_missing(name, *args)
+      start_time = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC)
+      result = @subject.public_send(name, *args)
+    ensure
+      end_time = ::Process.clock_gettime(::Process::CLOCK_MONOTONIC) - start_time
+      ::Kernel.puts "-- call #{name} args |#{args.inspect[0..100]}| response |#{result.inspect[0..100]}| - (#{end_time}s)"
+      result
+    end
+
+    def respond_to_missing?(method_name, include_private = false)
+      super
+    end
+  end
+end
--- a/lib/3scale_toolbox.rb
+++ b/lib/3scale_toolbox.rb
@@ -1,6 +1,7 @@
 require '3scale_toolbox/version'
 require '3scale_toolbox/helper'
 require '3scale_toolbox/error'
+require '3scale_toolbox/proxy_logger'
 require '3scale_toolbox/swagger'
 require '3scale_toolbox/configuration'
 require '3scale_toolbox/remotes'
--- a/lib/3scale_toolbox/3scale_client_factory.rb
+++ b/lib/3scale_toolbox/3scale_client_factory.rb
@@ -1,17 +1,18 @@
 module ThreeScaleToolbox
   class ThreeScaleClientFactory
     class << self
-      def get(remotes, remote_str, verify_ssl)
-        new(remotes, remote_str, verify_ssl).call
+      def get(remotes, remote_str, verify_ssl, verbose = false)
+        new(remotes, remote_str, verify_ssl, verbose).call
       end
     end
 
-    attr_reader :remotes, :remote_str, :verify_ssl
+    attr_reader :remotes, :remote_str, :verify_ssl, :verbose
 
-    def initialize(remotes, remote_str, verify_ssl)
+    def initialize(remotes, remote_str, verify_ssl, verbose)
       @remotes = remotes
       @remote_str = remote_str
       @verify_ssl = verify_ssl
+      @verbose = verbose
     end
 
     def call
@@ -21,7 +22,10 @@ module ThreeScaleToolbox
         remote = remotes.fetch(remote_str)
       end
 
-      remote_client(remote.merge(verify_ssl: verify_ssl))
+      client = remote_client(remote.merge(verify_ssl: verify_ssl))
+      return ProxyLogger.new(client) if verbose
+
+      client
     end
 
     private
--- a/lib/3scale_toolbox/base_command.rb
+++ b/lib/3scale_toolbox/base_command.rb
@@ -47,12 +47,16 @@ module ThreeScaleToolbox
     # Input param can be endpoint url or remote name
     #
     def threescale_client(str)
-      ThreeScaleClientFactory.get(remotes, str, verify_ssl)
+      ThreeScaleClientFactory.get(remotes, str, verify_ssl, verbose)
     end
 
     def verify_ssl
       # this is flag. It is either true or false. Cannot be nil
       !options[:insecure]
+    end
+
+    def verbose
+      options[:verbose]
     end
 
     def exit_with_message(message)
--- a/lib/3scale_toolbox/commands/3scale_command.rb
+++ b/lib/3scale_toolbox/commands/3scale_command.rb
@@ -19,6 +19,7 @@ module ThreeScaleToolbox
             exit 0
           end
           flag :k, :insecure, 'Proceed and operate even for server connections otherwise considered insecure'
+          flag nil, :verbose, 'Verbose mode'
           flag :h, :help, 'show help for this command' do |_, cmd|
             puts cmd.help
             exit 0
--- a/spec/unit/3scale_client_factory_spec.rb
+++ b/spec/unit/3scale_client_factory_spec.rb
@@ -6,24 +6,24 @@ RSpec.describe ThreeScaleToolbox::ThreeScaleClientFactory do
   let(:endpoint) { 'https://example.com' }
   let(:authentication) { '123456789' }
   let(:verify_ssl) { true }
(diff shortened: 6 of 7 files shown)


eguzki:
Another options to take into account would be to add a verbose flag in [3scale-api-ruby](https://github.com/3scale/3scale-api-ruby) to print HTTP request params and response status and body. Just food for thought

lib/3scale_toolbox/proxy_logger.rb
```diff
@@ -0,0 +1,21 @@
...
+      # skipping undef of methods that "may cause serious problems"
+      undef_method(meth) if meth !~ /^(__|object_id)/
+    end
+
+    def initialize(subject)
+      @subject = subject
+    end
+
+    def method_missing(name, *args)
+      start_time = Time.now
```
mikz: Better to use Benchmark module for this.

lib/3scale_toolbox/proxy_logger.rb
```diff
@@ -0,0 +1,21 @@
...
+      result = @subject.send(name, *args)
```
mikz: public_send?

lib/3scale_toolbox/proxy_logger.rb
```diff
@@ -0,0 +1,21 @@
...
+    instance_methods.each do |meth|
+      # skipping undef of methods that "may cause serious problems"
+      undef_method(meth) if meth !~ /^(__|object_id)/
+    end
+
+    def initialize(subject)
+      @subject = subject
+    end
+
+    def method_missing(name, *args)
```
mikz: When implementhing method_missing respond_to_missing should be also implemented.

lib/3scale_toolbox/proxy_logger.rb
```diff
@@ -0,0 +1,21 @@
+module ThreeScaleToolbox
+  class ProxyLogger
```
mikz: Why not using BasicObject ?

mikz: Or actual Delegator ?

lib/3scale_toolbox/proxy_logger.rb
eguzki: you mean something like this?? 
```ruby
def method_missing(name, *args)                                                                   
  result = nil                                                                                    
  time = Benchmark.realtime do                                                                    
    result = @subject.send(name, *args)                                                           
  end        
ensure                                                                                     
  puts "-- call #{name} args |#{args.inspect[0..80]}| response |#{result.inspect[0..80]}| - (#{time}s)"
end                                                                                               
```

I do not see any enhancement doing that. Besides, if error raised from subject this implementation does not print time elapsed, while current one does.

eguzki:
Updated `ProxyLogger`. Added `respond_to_missing?` and inherited from `BasicObject`.

`ProxyLogger.new(MyClass.new).method(:mymethod)`      => #<Method: MyClass#mymethod>
`ProxyLogger.new(MyClass.new).class`                              => MyClass

@mikz :+1:

lib/3scale_toolbox/proxy_logger.rb
mikz: The first enhancement is that Benchmark.realtime is measuring the time properly, using monotonic time.

If you care about the error case, then just use monotonic time.

lib/3scale_toolbox/proxy_logger.rb
```diff
@@ -0,0 +1,25 @@
+module ThreeScaleToolbox
+  class ProxyLogger < BasicObject
+    instance_methods.each do |meth|
```
mikz: There is no need to undefine those methods anymore no?

eguzki: There is no need to `undef`, right. Fixed.

Regarding Delegator/SimpleDelegator: I am trying to intercept calls, log input/output and measure elapsed time. This is done by proxy pattern. Decorator pattern using Delegator/SimpleDelegator is not meant for that, as long as I understood it.  Delegating is meant to add functionality in a class, but still behave like that. Besides solves inheritance issues to behave  in a way that's specific to delegated object, like `.class` method returned info.

mikz: Delegator forwards all messages to the delegated objects. If you want to intercept all of them, then just need to override method_missing in the same way you do here, except instead of actually delegating the message sending just call super.

Delegator implements methods for equality, hashing, comparison, etc. that might not be needed in this case. The implementation is trivial, so it is not even about writing less code.

I just feel it would be a bit more correct because delegating all the methods to some other object is the purpose of the SimpleDelegator as mentioned in the docs:
> A concrete implementation of Delegator, this class provides the means to delegate all supported method calls to the object passed into the constructor

Issue #91, opened 2019-02-05
RFE: Set the integration settings when importing a service from an OpenAPI Specification


nmasse-itix:
# Current Situation

Currently, when importing a new service from an OpenAPI with:

```sh
3scale import openapi -t echo_api -d https://$ACCESS_TOKEN@$TENANT-admin.3scale.net examples/echo-api.yaml
```

The service is created but the integration settings are left empty and the service is not ready to use. 

# Proposed change

Re-use the `host` and `schemes` fields of the OpenAPI Specification to set the private base URL. 
Re-use the standard `securityScheme` and `securityRequirements` fields to set the correct credentials location (query string or headers + header / parameter name). 

Also set the OpenID Connect issuer endpoint either from the OpenAPI Specification or from the command line. 

Ideally, those values should be taken from the OpenAPI Specification as default value that can be overridden from the command line. 

Option: accept a command line option to strip or replace the `basePath` field with custom value (hence setting the `url_rewritting` policy in the policy chain). 

# Business justification

If a service is left half-provisioned, the service is not usable and someone has to come finished the configuration manually or write a higher-level script.

sgutierr:
Setting the private url is foundamental for CICD pipeline. for instance setting a private url for each environment (non-prod, prod)

nmasse-itix:
Related to #47: when a service is imported from an OpenAPI Specification and the integration settings are defined, they must be defined for the staging env. 

The user would then have the ability to run smoke tests on the staging gateway by any means. 

Then, the user would need to promote the proxy settings from staging to production using a `3scale proxy promote` command (or equivalent).

eguzki:
@nmasse-itix Regarding configuring `url_rewritting` policy, I do not have clear idea about what you suggest.

Having optional parameter like `--base-path` to override base path from OAS spec is clear. But, configure  `url_rewritting` policy, how? adding `base path` prefix to all request paths? Take into account that when creating mapping rules, the toolbox is already appending base_path to OAS operations. What is the intended functionality?

Another doubt, configure url-rewritting policy only when `base-path` is overriden or always when importing OAS??

nmasse-itix:
Sure, let's start with an example. 

Let's say that our customer has an OAS with one method: `POST /say-hello`. The `basePath` in the OAS might or might not be present and might not even have the correct value. 

There are three elements we need to consider:

- what the customer would like to expose (ex: `/api1/say-hello`)
[...]

PR #116, opened 2019-03-20
Openapi import: update service settings based on security reqs

### LINKED ISSUE #92: RFE: Set the correct security scheme when importing an OpenAPI Specification
# Current Situation

Currently, when importing a new service from an OpenAPI with:

```sh
3scale import openapi -t echo_api -d https://$ACCESS_TOKEN@$TENANT-admin.3scale.net examples/echo-api-oidc.yaml
```

The service is created with API Key security even if the OpenAPI Specification mandates OpenID Connect and I found no specific way to override this from the command line. 

See this sample OpenAPI Specification: [echo-api-oidc.yaml](https://github.com/nmasse-itix/threescale-cicd/blob/f22a0fc3cb70f91cc6edc11117bb14f3940b5730/tests/test-cases/api-contracts/echo-api-oidc.yaml#L29-39). 

# Proposed Change

The 3scale toolbox looks for the `securityRequirements` and `securitySchemes` fields to determine the kind of security to apply and sets it accordingly when creating or updating the service.

The current logic that I'm using in the Ansible playbooks are:
 - I'm looking for the global security requirements and I require to have one and only one requirement
 - I'm searching for the matching security scheme
 - I'm determining whether the scheme is API Key or OIDC (the two we are supporting in 3scale)

If there is no security requirement, we can consider it as an "Open API" and add the [default_credentials policy](https://github.com/3scale/apicast/tree/master/gateway/src/apicast/policy/default_credentials) to the policy chain. 

# Business Justification

One of our customer is using exclusively OpenID Connect APIs and thus cannot use the 3scale toolbox to provision its APIs.

eguzki:
Set the correct service authentication settings when importing an OpenAPI Specification based on security requirements.

- [x] Update swagger parser to parse security requirements and definitions (schemes)
- [x] Update service api_backend (private base URL)
- [x] Update service auth mode (backend_version)
- [x] Update service auth settings for apiKey mode
- [x] Update service auth settings for oidc mode
- [x] Update service auth settings when no auth is specified
- [x] tests

Fixes #92 

Partially implementing #91, but not fully, just setting `private base URL`

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
@@ -1,0 +1,84 @@
+module ThreeScaleToolbox
+  module Commands
+    module ImportCommand
+      module OpenAPI
+        class UpdatePoliciesStep
+          include Step
+
+          ##
+          # Updates Proxy Policies config
+          def call
+            # to be idempotent, we must read first
+            source_policies_settings = service.policies
+            # shallow copy
+            # to detect policies_settings changes, should only be updated setting objects
+            # do not update in-place, otherwise changes will not be detected
+            policies_settings = source_policies_settings.dup
+
+            add_anonymous_access_policy(policies_settings)
+            add_rh_sso_keycloak_role_check_policy(policies_settings)
+
+            return if source_policies_settings == policies_settings
+
+            service.update_policies('policies_config' => policies_settings)
+          end
+
+          private
+
+          def add_anonymous_access_policy(policies)
+            # only on 'open api' security req
+            return unless security.nil?
+
+            return if policies.any? { |policy| policy['name'] == 'default_credentials' }
+
+            # Anonymous policy should be before apicast policy
+            # hence, adding as a first element
+            policies.insert(0, anonymous_policy)
+          end
+
+          def anonymous_policy
+            raise ThreeScaleToolbox::Error, 'User key must be provided by ' \
+              '--default-credentials-userkey optional param' \
+              if default_credentials_userkey.nil?
+
+            {
+              'name': 'default_credentials',
+              'version': 'builtin',
+              'configuration': {
+                'auth_type': 'user_key',
+                'user_key': default_credentials_userkey
+              },
+              'enabled': true
+            }
+          end
+
+          def add_rh_sso_keycloak_role_check_policy(policies)
+            # only applies to oauth2 sec type
+            return if security.nil? || security.type != 'oauth2'
+
+            return if policies.any? { |policy| policy['name'] == 'keycloak_role_check' }
+
+            policies << keycloak_policy
+          end
+
+          def keycloak_policy
+            {
+              'name': 'keycloak_role_check',
+              'version': 'builtin',
+              'configuration': {
+                'type': 'whitelist',
+                'scopes': [
+                  {
+                    'realm_roles': [],
+                    'client_roles': security.scopes.map { |scope| { 'name': scope } }
+                  }
+                ]
+              },
+              'enabled': true
+            }
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
@@ -1,0 +1,55 @@
+module ThreeScaleToolbox
+  module Commands
+    module ImportCommand
+      module OpenAPI
+        class UpdateServiceOidcConfStep
+          include Step
+
+          ##
+          # Updates OIDC config
+          def call
+            # setting required attrs, operation is idempotent
+            oidc_settings = {}
+
+            add_flow_settings(oidc_settings)
+
+            return unless oidc_settings.size.positive?
+
+            service.update_oidc oidc_settings
+            puts 'Service oidc updated'
+          end
+
+          private
+
+          def add_flow_settings(settings)
+            # only applies to oauth2 sec type
+            return if security.nil? || security.type != 'oauth2'
+
+            oidc_configuration = {
+              standard_flow_enabled: false,
+              implicit_flow_enabled: false,
+              service_accounts_enabled: false,
+              direct_access_grants_enabled: false
+            }.merge(flow => true)
+            settings.merge!(oidc_configuration)
+          end
+
+          def flow
+            case (flow_f = security.flow)
+            when 'implicit'
+              :implicit_flow_enabled
+            when 'password'
+              :direct_access_grants_enabled
+            when 'application'
+              :service_accounts_enabled
+            when 'accessCode'
+              :standard_flow_enabled
+            else
+              raise ThreeScaleToolbox::Error, "Unexpected security flow field #{flow_f}"
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
@@ -1,0 +1,62 @@
+module ThreeScaleToolbox
+  module Commands
+    module ImportCommand
+      module OpenAPI
+        class UpdateServiceProxyStep
+          include Step
+
+          ##
+          # Updates Proxy config
+          def call
+            # setting required attrs, operation is idempotent
+            proxy_settings = {}
+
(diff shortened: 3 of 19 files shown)


eguzki:
Ready for review @mikz @miguelsorianod

lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
```diff
@@ -0,0 +1,92 @@
...
+            service.update_policies('policies_config' => policies_settings)
+          end
+
+          private
+
+          def add_anonymous_access_policy(policies)
+            # only on 'open api' security req
+            return unless security.nil?
+
+            def_credentials = policies.find { |policy| policy['name'] == 'default_credentials' }
```
mikz: `policies.any?`

lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
```diff
@@ -0,0 +1,92 @@
...
+            }
+          end
+
+          def add_rh_sso_keycloak_role_check_policy(policies)
+            # only applies to oauth2 sec type
+            return if security.nil? || security.type != 'oauth2'
+
+            keycloak_policy_idx = policies.find_index { |policy| policy['name'] == 'keycloak_role_check' }
+
+            return unless keycloak_policy_idx.nil?
```
mikz: why find_index ? isn't `.any?` enough?

lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
```diff
@@ -0,0 +1,92 @@
...
+            source_policies_settings = service.policies
+            # shallow copy
+            # to detect policies_settings changes, should only be updated setting objects
+            # do not update in-place, otherwise changes will not be detected
+            policies_settings = source_policies_settings.dup
+
+            add_anonymous_access_policy(policies_settings)
+            add_rh_sso_keycloak_role_check_policy(policies_settings)
+
+            return if same_policy_settings?(source_policies_settings, policies_settings)
```
mikz: This seems a bit weird to extract into a method.

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
```diff
@@ -0,0 +1,67 @@
...
+              settings[:credentials_location] = 'headers'
+              settings[:oidc_issuer_endpoint] = oidc_issuer_endpoint unless oidc_issuer_endpoint.nil?
+            when 'apiKey'
+              settings[:credentials_location] = credentials_location
+              settings[:auth_user_key] = security.name
+            else
+              raise ThreeScaleToolbox::Error, "Unexpected security scheme type #{security.type}"
+            end
+
+            nil
```
mikz: Why retururning nil?

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
```diff
@@ -0,0 +1,67 @@
...
+          end
+
+          def credentials_location
+            return nil if security.nil?
```
mikz: Why return nil over just return?

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
```diff
@@ -0,0 +1,67 @@
...
+
+            case security.in_f
+            when 'query'
+              'query'
+            when 'header'
+              'headers'
+            else
+              raise ThreeScaleToolbox::Error, "Unexpected security in_f field #{security.in_f}"
```
mikz: This is a potentially untested branch with a risk of having a typo in the interpolation. Using variables would prevent that `case (in_f = security.in_f)`.

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
```diff
@@ -0,0 +1,67 @@
...
+          end
+
+          private
+
+          def add_api_backend_settings(settings)
+            scheme = api_spec.schemes.first || 'https'
+            host = api_spec.host
+
+            settings[:api_backend] = "#{scheme}://#{host}"
+            nil
```
mikz: Why returning nil?

lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
eguzki: yeah, you are right. Long history short is legacy code from old implementation.

lib/3scale_toolbox/commands/import_command/openapi/update_policies_step.rb
eguzki: Right. Initially deep compare was implemented. Later I changed strategy and only shallow copies and shallow compares are used.

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
eguzki: no reason. :woman_shrugging:

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
eguzki: no reason :woman_shrugging:

eguzki: actually, `security.nil?` can never happen. `credentials_location` is only used in a method that  checks previously that. I was just adding safety checking just in case that method is moved somewhere else.

eguzki: imaginary problems

mikz: Well, my question was more about why `return nil if security.nil?` over `return unless security`, not why the check at all.

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
mikz: Then it does not need to exist, right? :)

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
eguzki: Adding test for that branch

mikz: My point wasn't really about this one branch, but general about not calling a method in a less used branch and rather using a variable.

eguzki:
Fixed all comments. Thnx @mikz

lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
```diff
@@ -0,0 +1,57 @@
...
+            when 'implicit'
+              :implicit_flow_enabled
+            when 'password'
+              :direct_access_grants_enabled
+            when 'application'
+              :service_accounts_enabled
+            when 'accessCode'
+              :standard_flow_enabled
+            else
+              raise ThreeScaleToolbox::Error, "Unexpected security flow field #{security.flow}"
```
mikz: I haven't commented on this, but it is the same case as the other rescue branch.

lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
```diff
@@ -0,0 +1,57 @@
...
+              standard_flow_enabled: false,
+              implicit_flow_enabled: false,
+              service_accounts_enabled: false,
+              direct_access_grants_enabled: false
+            }.merge(flow => true)
+            settings.merge!(oidc_configuration)
+          end
+
+          def flow
+            return nil if security.nil?
```
mikz: I'm usually not commenting on all instances of one issue I find, so this one stayed.

lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
eguzki: :+1:

lib/3scale_toolbox/commands/import_command/openapi/update_service_oidc_conf_step.rb
eguzki: reviewed all the code for issues like this and it seems that all of them fixed.

PR #120, opened 2019-03-27
Swagger parser: whitelist for  path object elements

eguzki:
- [x] Swagger parser: whitelist for  path object elements
- [x] enhance `system_name` sanitize method
- [x] catch errors on several api calls
- [x] add oidc service import integration test

### DIFF
--- a/spec/unit/swagger/swagger_spec.rb
+++ b/spec/unit/swagger/swagger_spec.rb
@@ -194,6 +194,78 @@ RSpec.describe ThreeScaleToolbox::Swagger do
         expect(get_findPetsByStatus_operation.operation_id).to eq('findPetsByStatus')
       end
     end
+
+    context 'extensions in path object' do
+      let(:content) do
+        <<~YAML
+          ---
+          swagger: "2.0"
+          info:
+            title: "#{title}"
+            version: "1.0.0"
+          paths:
+            /pet:
+              get:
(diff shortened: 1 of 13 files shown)


spec/unit/swagger/swagger_spec.rb
```diff
@@ -194,6 +194,78 @@
...
+
+      it 'parsed operation path /pet' do
+        expect(subject.operations[0].path).to eq('/pet')
+      end
+      it 'parsed operation method get' do
+        expect(subject.operations[0].verb).to eq('get')
+      end
+    end
+
+    context 'parameters in path item object' do
```
miguelsorianod: Would it be a good idea to test that there are parameters?

eguzki: `parameters` fields are not parsed, they are filtered out. The context says there are parameters. If there are no parameters fields in the `content`, the test is useless.

To test that there are `parameters`, you could parse `content` and make sure parameters are there. but I think it is so obvious because there is a specific context for that case

eguzki: In other words, if there is a context for some var `A` to be, let's say, `2`, I think it is so obvious that there is no need to check that `A` is indeed `2`

miguelsorianod: :+1:

PR #121, opened 2019-03-27
Import openapi: update url-rewritting policy

eguzki:
Implements spec from #91 

Specifically https://github.com/3scale/3scale_toolbox/issues/91#issuecomment-475641632

Adding more integration tests on importing OAS was getting complicated to keep up with single mocked results generic for all test. So I ended up putting each integration test on a different file and share common context. Little refactor was required, but I guess the resulting testing code allows adding more e2e tests about importing OAS.

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi.rb
@@ -33,6 +33,8 @@ module ThreeScaleToolbox
               flag    nil, 'skip-openapi-validation', 'Skip OpenAPI schema validation'
               option  nil, 'oidc-issuer-endpoint', 'OIDC Issuer Endpoint', argument: :required
               option  nil, 'default-credentials-userkey', 'Default credentials policy userkey', argument: :required
+              option  nil, 'override-private-basepath', 'Override the basepath for the public URLs', argument: :required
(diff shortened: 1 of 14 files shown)


eguzki:
21 days opened already. @mikz @miguelsorianod 

I have a bot installed that on 25th day, it is merged automatically ;) :trollface:

lib/3scale_toolbox/commands/import_command/openapi.rb
```diff
@@ -33,6 +33,8 @@ def self.command
               flag    nil, 'skip-openapi-validation', 'Skip OpenAPI schema validation'
               option  nil, 'oidc-issuer-endpoint', 'OIDC Issuer Endpoint', argument: :required
               option  nil, 'default-credentials-userkey', 'Default credentials policy userkey', argument: :required
+              option  nil, 'override-private-basepath', 'Override the basepath for the public URLs', argument: :required
```
mikz: I could understand this as that I can override the whole base URL. But the tests show it is only the path component.

But I don't have enough understanding of the topic.

eguzki: All  the information is in the issue #91 

Your are, basically, overriding [basePath](https://swagger.io/specification/v2/) field from the swagger spec and configuring url-rewritting policy accordingly

PR #124, opened 2019-04-03
Import openapi: patch activedocs

### LINKED ISSUE #90: RFE: Create an ActiveDocs with updated environmental fields when importing a service from OpenAPI

eguzki:
Fixes #90 

OpenAPI Specification is patched to take into account the new environment:
* the `host` field is updated with the hostname component of the public production base URL
* the `schemes` field is updated with the scheme component of the public production base URL
* the `basePath` field is updated if rewriting rules have been set
* the `securityDefinitions/*/authorizationUrl` and `securityDefinitions/*/tokenUrl` fields are updated with the token and authorize endpoints listed in the OpenID Connect wellknown endpoint (when OIDC is in use)


eguzki:
@miguelsorianod PR ready for review

PR #126, opened 2019-04-05
Fix copying self managed service to on prem where invalid

eguzki:
When copying a service from 3scale SaaS to an On-premises deployment using the toolbox command and APIcast is configured as self-managed gateway: 
```
3scale copy service 12345678 --source=https://<SaaS_provider_key>@<SaaSdomain>-admin.3scale.net --destination=https://<OnPremprovider_key>@3scale-admin.example.com --target_system_name=abcdefg
```

It outputs the following error: 
```
Error: Service has not been saved. Errors: {"deployment_option"=>["is not included in the list"]}
```

The problem is that 3scale On-premises  doesn't have the self-managed gateway option in the configuration.

### DIFF
--- a/lib/3scale_toolbox/tasks/update_service_settings_task.rb
+++ b/lib/3scale_toolbox/tasks/update_service_settings_task.rb
@@ -11,13 +11,30 @@ module ThreeScaleToolbox
 
       def call
         source_obj = source.show_service
-        response = target.update_service(target_service_params(source_obj))
-        raise ThreeScaleToolbox::Error, "Service has not been saved. Errors: #{response['errors']}" unless response['errors'].nil?
+        svc_obj = update_service target_service_params(source_obj)
+        if (errors = svc_obj['errors'])
+          raise ThreeScaleToolbox::Error, "Service has not been saved. Errors: #{errors}" \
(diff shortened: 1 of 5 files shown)


eguzki:

lib/3scale_toolbox/tasks/update_service_settings_task.rb
```diff
@@ -11,14 +11,30 @@ def initialize(source:, target:, target_name:)
...
       private
 
+      def update_service(service)
+        svc_obj = target.update_service service
+
+        # Invalid deployment option check
+        # use default deployment_option
+        if (errors = svc_obj['errors']) &&
+           ThreeScaleToolbox::Helper.service_invalid_deployment_option?(errors)
+          service.delete('deployment_option')
```
miguelsorianod: I think might be good to clarify why the deployment_option is deleted and then the update retried.
It's hard to know why is being done for that parameter only and not for others

eguzki: :+1:

eguzki:
Addded clarification on invalid deployment option check

PR #129, opened 2019-04-10
proxy update as last task to update proxy object with latest changes

eguzki:
**Current situation**
When updating/creating proxy objects (the mapping rules, policies, oidc conf, auth conf, public base URLs), then this requires a second, not very intuitive, API call (i.e. proxy update) to update the proxy object and bump proxy config version.

If the proxy update call is not done, it can be very confusing because the mapping rules will appear in the integration page but they will not appear in the proxy config object. The same applies to mapping rules API calls. They changes will be available on mapping rules endpoints, but not available to the gateway because the proxy proxy config object has not changed.

This behavior is intended. In the API case you need to make several API calls and then bump the version. 

**Proposed changes**
This PR re-orders command tasks to make sure proxy update call is done *after* all updates are done. Proxy object should be updated with the latest changes made by the toolbox in `staging` environment.

Proxy object changes should not be applied to `production` environment. This task should be performed by a separate promoting command

### DIFF
--- a/lib/3scale_toolbox/commands/copy_command/copy_service.rb
+++ b/lib/3scale_toolbox/commands/copy_command/copy_service.rb
@@ -33,7 +33,6 @@ module ThreeScaleToolbox
           puts "new service id #{target_service.id}"
           context = create_context(source_service, target_service)
           tasks = [
-            Tasks::CopyServiceProxyTask.new(context),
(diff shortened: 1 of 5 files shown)


eguzki:
Feature ready to review @miguelsorianod

PR #132, opened 2019-04-11
Export/import Application Plan limits, pricing rules and features

eguzki:
A new command `application-plan` is introduced to export/import application plan artefacts from/to a YAML file.

Export
```
NAME
    export - export application plan

USAGE
    3scale application-plan export [opts] <remote> <service_id_or_system_name> <plan_system_name>

DESCRIPTION
    Export application plan, including limits, pricing rules and features

OPTIONS
    -f --file=<value>             Write to file instead of stdout
```
Import 
```
NAME
    import - import application plan

USAGE
    3scale application-plan import [opts] <remote>
    <service_id_or_system_name>

DESCRIPTION
    Import application plan, including limits, pricing rules and features

OPTIONS
    -f --file=<value>             Read from file instead of stdin
    -p --plan-system-name=<value>      Override application plan system name
```

Fixes #94 
Fixes #65

PR #134, opened 2019-05-02
[WIP] Application plan CRUD operations

### LINKED ISSUE #41: Command to create, list, show, and delete application plans
Lower level command for application plans management

eguzki:
* Create application plan
```
NAME
    create - create application plan

USAGE
    3scale application-plan create [opts] <remote>
    <service> <plan_name>

DESCRIPTION
    Create application plan

OPTIONS
       --approval-required              Applications require approval
       --cost-per-month=<value>         Cost per month
    -d --default                        Make default application plan
       --disabled                       Disable application plan
       --end-user-required              End user required
    -p --published                      Publish application plan
       --setup-fee=<value>              Setup fee
    -t --system-name=<value>            Application plan system name
       --trial-period-days=<value>      Trial period days

OPTIONS FOR APPLICATION-PLAN
    -c --config-file=<value>            3scale toolbox configuration file
                                        (default:
                                        /home/eguzki/.3scalerc.yaml)
    -h --help                           show help for this command
    -k --insecure                       Proceed and operate even for server
                                        connections otherwise considered
                                        insecure
    -v --version                        Prints the version of this command
       --verbose                        Verbose mode
```

* List application plans (TBD)

* Show application plan (TBD)

* Apply (update if it exits, otherwise create) application plan (TBD)

* Delete application plans (TBD)

Fixes #41


eguzki:
@thomasmaas it would be good if you review command syntax, naming, etc... Also documentation for mistakes, nitpicks

eguzki:

Export/import Application Plan limits, pricing rules and features #132 should be merged first. This PR needs some modules built in that PR #132

docs/app-plan.doc
```diff
@@ -0,0 +1,173 @@
...
+
+DESCRIPTION
+    Update (create if it does not exist) application plan
+
+OPTIONS
+       --approval-required=<value>      Applications require approval. true
+                                        or false
+       --cost-per-month=<value>         Cost per month
+       --default                        Make default application plan
+       --disabled                       Disable application plan
```
thomasmaas: Disables all limits for this Application Plan

docs/app-plan.doc
```diff
@@ -0,0 +1,173 @@
...
+       --enabled                        Enable application plan
```
thomasmaas: I wonder if this is needed. It would remove all 0/eternity limits I guess?

I was expecting a boolean: disabled: true false

docs/app-plan.doc
eguzki: :+1:

docs/app-plan.doc
eguzki: This is `apply` command, like `update`. The default behavior should be only update things I tell you to update. If we implement `--disabled` default value as `false`, hence enabled, updating disabled app plan to change only, for instance  the `cost-per-month`, would enable as well the application plan, deleting all eternity zero metrics.

For the `create` command, it makes sense your expectation and that is the behavior it was implemented for the `create` command.

IMO, for the `apply` command, you should explicitly state you want the plan to be `enabled` or `disabled` or do not change current status.

The same principle applies for `approval-required` and `end-user-required`. This is why, they are not `flags`, but `options` with required value. If not set, `apply` will not change their current state. And if you want to update them, you must provide a value, either `true` or `false`

WDYT?? make this sense to you?

miguelsorianod:
Hi,

I think it would be useful to see whether the application-plan is published or not via the CLI when using the show command

thomasmaas:
we can leave that for a future update @miguelsorianod. Let's merge!

miguelsorianod:
I already accepted the PR before :+1: and I was asked to do additional manual testing from @eguzki :).

The testings I've performed look good. To me can be merged.

PR #142, opened 2019-05-16
[WIP] Add find and find accounts command

orimarti:
Added find super command and find accounts command.

Find account:
* Created class account with self method to find accounts by email, provider_key or service_token
* Class accounts have the to_s method, so only printing it will print all the info
* Find account command that calls find method and prints it


eguzki:
Latest convention for commands is `subject action`, so it should be:

`3scale account find`

currently you are implementing:

`3scale find account`

lib/3scale_toolbox/commands/find_command/find_account.rb
```diff
@@ -0,0 +1,39 @@
...
+            option :i, :instance, '3scale instance. Url or remote name', argument: :required
+            option :a, :'print-all', 'Print all the account info', argument: :forbidden
+            param :text
+
+            runner FindAccountSubcommand
+          end
+        end
+
+        def run
+          client = threescale_client(options[:instance])
```
eguzki: instance is an optional param. What if it is not given in the command? I would change it to positional param and therefore enforced by the CRI lib to exist

lib/3scale_toolbox/entities/base_entity.rb
```diff
@@ -0,0 +1,85 @@
...
+      ].freeze
+      DEFAULT_EXTRA_VARS_NAME = "extra_vars"
+      public_constant :DEFAULT_PRINTABLE_VARS
+      public_constant :DEFAULT_VERBOSE_PRINTABLE_VARS
+      public_constant :DEFAULT_EXTRA_VARS_NAME
+
+      attr_accessor :verbose
+
+      def self.included(base)
+        unless base.const_defined?(:PRINTABLE_VARS)
```
eguzki: I think we can workout this to make simpler.

This is what I would do, but it is open to discussion.
Remove `self.included` method. Entity has `PRINTABLE_VARS` public constant

```
module Entity                            
  PRINTABLE_VARS = %w[                   
        id                               
  ].freeze                               
  public_constant :PRINTABLE_VARS        
                                         
  VERBOSE_PRINTABLE_VARS = %w[           
        id                               
  ].freeze                               
  public_constant :VERBOSE_PRINTABLE_VARS

```

Entities, like Account, may or may not have those configured. But Entities module will have the following method to read printable attributes that will try to read child entity printable attributes and fallback to default defined in Entities module.

```
def printable_attributes                                           
  if verbose                                                       
    self.class.const_get(:VERBOSE_PRINTABLE_VARS, inherit: true)   
  else                                                             
    self.class.const_get(:PRINTABLE_VARS, inherit: true)           
  end                                                              
end                                                                

```

lib/3scale_toolbox/entities/base_entity.rb
```diff
@@ -0,0 +1,85 @@
...
+        if @verbose
+          format_vars(printable: self.class.const_get(:VERBOSE_PRINTABLE_VARS))
+        else
+          format_vars(printable: self.class.const_get(:PRINTABLE_VARS))
+        end
+      end
+
+      private
+
+      def format_vars (printable: nil)
```
eguzki: This method has some issues.
* it is recursive. Although not bad for all cases, IMO, adds complexity where there is not
* Iterates over `instance_variables`. It is way better to iterate over printable attributes because you already have that list.
* the algorithm considers attributes that can be arrays, hashes or plain types. I think there is no need. We can assume that entitie's attributes are all plain attributes. In the future, in we need it, we can increase complexity.

I have done a PoC, so you can see and run locally, just ruby plain script

```ruby
module Entity
  PRINTABLE_VARS = %w[
        id
  ].freeze
  public_constant :PRINTABLE_VARS

  VERBOSE_PRINTABLE_VARS = %w[
        id
  ].freeze
  public_constant :VERBOSE_PRINTABLE_VARS

  def printable_attributes
    if verbose
      self.class.const_get(:VERBOSE_PRINTABLE_VARS, inherit: true)
    else
      self.class.const_get(:PRINTABLE_VARS, inherit: true)
    end
  end

  def to_s
    attributes = attrs.merge('id' => id)
    value_array = printable_attributes.map do |attr|
      "#{attr} => #{attributes.fetch(attr, '(empty)')}"
    end

    value_array.join("\n")
  end

  module ClassMethods
  end
end


class Account
  include Entity

  VERBOSE_PRINTABLE_VARS = %w[id org_name email username].freeze
  public_constant :VERBOSE_PRINTABLE_VARS

  attr_reader :id, :attrs, :verbose
  def initialize(id:, attrs:, verbose: false)
    @id = id
    @attrs = attrs
    @verbose = verbose
  end
end

a = Account.new(id: 1, attrs: { 'org_name' => 'Some Corp' }, verbose: true)
puts a

puts '--------------------------'

b = Account.new(id: 2, attrs: { 'org_name' => 'Some Corp2' })
puts b
```

IMO it is simpler and works the same for most of the use cases.

Running it you get:

```
$ ruby entities_poc.rb 
id => 1
org_name => Some Corp
email => (empty)
username => (empty)
--------------------------
id => 2
```

PR #150, opened 2019-05-29
3scale import openapi set the public staging prod URLs

eguzki:
the "3scale import openapi" command has to set the public staging/prod URLs
==================================================================

When importing an OpenAPI, the service is created using the SaaS APIcast but there is no mean to specify a custom public staging/prod URL so that I could use my own APIcast instances.

The "3scale import openapi" command has two more optional parameters :
```
--staging-public-base-url=http://my-apicast-instance.mydomain.com:8080
--production-public-base-url=https://api.example.com
```


eguzki:
ready to be reviewed @miguelsorianod

PR #153, opened 2019-06-03
Copy policy registry

eguzki:
Toolbox command to `copy` policy registry (a.k.a. custom policies) from source tenant to target tenant.
Policy registry identificator is `#{name}-#{version}`. Having this identifying function, we can define
* Missing policies: policies that exist in source tenant and do not exist in target tenant. 
* Matching policies: policies from source tenant that exist in target tenant.

Then:
- [x] Missing policies are being created in target tenant
- [x] Matching policies are being updated in target tenant

This behavior makes the command idempotent.

Integration tests not implemented as they require two different provider accounts.


eguzki:

PR #199, opened 2019-07-05
ApplicationPlan, Method, Metric and Application integration test

eguzki:
* application plan apply command
* application plan create command
* application plan list command
* application plan delete command
* application plan show command
* method create command
* method apply command
* method list command
* method delete command
* metric apply command
* metric create command
* metric delete command
* metric list command
* application list command
* application create command
* application show command
* application apply command
* application delete command

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,16 @@ workflows:
       - ruby-262-unittest
       - license_check
       - help_test
+
+  integration:
+    jobs:
+      - manual_approval: # <<< A job that will require manual approval in the CircleCI web application.
+          type: approval # <<< This key-value pair will set your workflow to a status of "On Hold"
+          # On approval of the `hold` job, any successive job that requires the `hold` job will run.
+      - integration_tests:
+          context: autotestaccount
+          requires:
+            - manual_approval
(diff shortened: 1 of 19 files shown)


eguzki:
@miguelsorianod ready for review

.circleci/config.yml
```diff
@@ -96,6 +96,16 @@ workflows:
       - ruby-262-unittest
       - license_check
       - help_test
+
+  integration:
+    jobs:
+      - manual_approval: # <<< A job that will require manual approval in the CircleCI web application.
+          type: approval # <<< This key-value pair will set your workflow to a status of "On Hold"
+          # On approval of the `hold` job, any successive job that requires the `hold` job will run.
+      - integration_tests:
```
miguelsorianod: What would be the idea of this during each PR?
If I have understood correctly what this does is that it requires the user interaction in CircleCI to approve the step. Would this be left as an optional job in GitHub? what would be the criteria?

eguzki: Integration tests are run nightly.

With this new `integration` workflow, each PR integration tests could be run, always manually approved. Useful if you suspect something could be broken

PR #204, opened 2019-07-24
Update copy service documentation to clarify policies copy prerequisites

miguelsorianod:


vramosp:
The wording seems confusing, I think it'd be more clear if we say something like this is how you copy a service and its custom policies:
- step 1: copy the custom policies with command x
- step 2: copy the service with command y
The point would be to explain that it's 2 different commands and that you need to copy the custom policies first.

miguelsorianod:
Hi @vramosp,

This PR was already merged.

Take into account that this information has been written in the context of the copy service command so it was described from the service copy centric point of view. The command to copy  custom policies is another command and the list of available commands and what they can do is available in the main README.md that references the specific documentation of all of them

vramosp:
My point is here the user won't know it can be copied as well (with a diff command) and it'd be useful to know. What about just referencing the other command doc at the end, something like:

"Make sure that their respective custom policy definitions already exist in the destination where the service is to be copied before performing the copy. To learn more about copying custom policies check out the documentation (link)"

miguelsorianod:
Seems good to me. I'll open a PR with that new change.

Thank you.

PR #221, opened 2019-10-30
Add custom host header and secret token options for import openapi

miguelsorianod:
Add new options in the import openapi command to allow specifying
a custom host header and a secret token that the API gateway will
send to the configured backend API in the proxy configuration.

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
@@ -40,9 +40,9 @@ module ThreeScaleToolbox
           end
 
           def add_api_backend_settings(settings)
-            return if private_base_url.nil?
-
-            settings[:api_backend] = private_base_url
+            settings[:api_backend] = private_base_url if !private_base_url.nil?
+            settings[:secret_token] = backend_api_secret_token if !backend_api_secret_token.nil?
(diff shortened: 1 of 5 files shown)


lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
```diff
@@ -40,9 +40,18 @@ def add_sandbox_endpoint_settings(settings)
           end
 
           def add_api_backend_settings(settings)
-            return if private_base_url.nil?
+            if !private_base_url.nil?
```
eguzki: rubocop whistle about https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfUnlessModifier

eguzki:
looking good

eguzki:
The smarter way would be to parse from openapi spec. Happy to hear about how to accomodate api backend auth and host headers in swagger (2.0) spec.

lib/3scale_toolbox/commands/import_command/openapi/update_service_proxy_step.rb
miguelsorianod: Changed :+1:

miguelsorianod:
Ready for review @eguzki 

I've added some unit and integration tests.

PR #224, opened 2019-11-18
Add prefix matching flag for mapping rules in import openapi command

miguelsorianod:
Add --prefix-matching in import openapi command.
Pending tests. Tests will be added in this PR once the implemented approach is deemed correct.

### DIFF
--- a/lib/3scale_toolbox/commands/import_command/openapi.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi.rb
@@ -30,6 +30,7 @@ module ThreeScaleToolbox
               option  :t, 'target_system_name', 'Target system name', argument: :required
               flag    nil, 'activedocs-hidden', 'Create ActiveDocs in hidden state'
               flag    nil, 'skip-openapi-validation', 'Skip OpenAPI schema validation'
+              flag    nil, 'prefix-matching', 'Use prefix matching instead of strict matching on mapping rules derived from openapi operations'
               option  nil, 'oidc-issuer-endpoint', 'OIDC Issuer Endpoint', argument: :required
               option  nil, 'default-credentials-userkey', 'Default credentials policy userkey', argument: :required
               option  nil, 'override-private-basepath', 'Override the basepath for the private URLs', argument: :required
@@ -74,7 +75,7 @@ module ThreeScaleToolbox
             openapi_resource = load_resource(arguments[:openapi_resource])
             {
               api_spec_resource: openapi_resource,
-              api_spec: ThreeScaleApiSpec.new(load_openapi(openapi_resource), options[:'override-public-basepath']),
+              api_spec: ThreeScaleApiSpec.new(load_openapi(openapi_resource), options[:'override-public-basepath'], options[:'prefix-matching']),
               threescale_client: threescale_client(fetch_required_option(:destination)),
               target_system_name: options[:target_system_name],
               activedocs_published: !options[:'activedocs-hidden'],
--- a/lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
+++ b/lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
@@ -17,8 +17,9 @@ module ThreeScaleToolbox
           end
 
           def pattern
-            # apply strict matching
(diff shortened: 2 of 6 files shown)


lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
```diff
@@ -17,8 +17,9 @@ def http_method
           end
 
           def pattern
-            # apply strict matching
-            "#{raw_pattern}$"
+            res = "#{raw_pattern}"
+            res = "#{res}$" if !operation[:prefix_matching] # apply strict matching
```
miguelsorianod: I implemented the setting of the pattern depending on the prefix_matching as an attribute of the Operation class.
However I'm not really sure of whether it would be the correct approach depending on what's the design of the separation of concerns among different classes I've seen

Possible solution candidates I've seen are:
 * Send the flag as a parameter of the ThreeScaleApiSpec constructor and also send it as part of the operation hash passed to the Operation class from the "operations" method in ThreeScaleApiSpec as done now.
 * Set the flag as part of the step context and as part of the CreateMappingRule Step modify the returned operations using the set method
 * Set the flag as part of the Swagger.build method. This does not seem really correct as I understand that class should be independent of 3scale specifics

What do you think? Where should the responsibilities be delimited?

lib/3scale_toolbox/commands/import_command/openapi/threescale_api_spec.rb
```diff
@@ -5,9 +5,10 @@ module OpenAPI
         class ThreeScaleApiSpec
           attr_reader :openapi
 
-          def initialize(openapi, base_path = nil)
+          def initialize(openapi, base_path = nil, prefix_matching = false)
```
miguelsorianod: In case this approach is taken, is it a good idea to have two default parameters? If we were to have the case where we want to set prefix_matching but not base_path(we want default for it) making this would force us to define the base_path parameter explicitly as nil.

lib/3scale_toolbox/commands/import_command/openapi.rb
```diff
@@ -30,6 +30,7 @@ def self.command
               option  :t, 'target_system_name', 'Target system name', argument: :required
               flag    nil, 'activedocs-hidden', 'Create ActiveDocs in hidden state'
               flag    nil, 'skip-openapi-validation', 'Skip OpenAPI schema validation'
+              flag    nil, 'prefix-matching', 'Use prefix matching instead of full path matching on mapping rules derived from openapi operations'
```
eguzki: instead of "full path matching", "strict matching" naming is widely used

lib/3scale_toolbox/commands/import_command/openapi/mapping_rule.rb
eguzki: I agree with the analysis. Options 1) and 2) are the best. Third one does not seem correct.

I would also go for the 1) option

lib/3scale_toolbox/commands/import_command/openapi/threescale_api_spec.rb
eguzki: I think it is ok.

lib/3scale_toolbox/commands/import_command/openapi/threescale_api_spec.rb
```diff
@@ -44,6 +45,10 @@ def security
             @security ||= parse_security
           end
 
+          def prefix_matching
```
eguzki: this is exactly what ` attr_reader :prefix_matching`  does

lib/3scale_toolbox/commands/import_command/openapi.rb
miguelsorianod: :+1:

lib/3scale_toolbox/commands/import_command/openapi/threescale_api_spec.rb
miguelsorianod: :+1:

miguelsorianod:
@eguzki requested changes applied and added some tests. Ready for review.

PR #225, opened 2019-11-19
Allow filtering activedocs by Service ID

miguelsorianod:
Implements activedocs list filtering by Service ID.

The original request about this https://issues.jboss.org/browse/THREESCALE-2967 suggested accepting service system name too.
For the moment I've implemented filtering by service ID due to:
 * Implement filtering by service system name would require performing an additional API call to system to get the associated service id, as the output from ActiveDocs list only returns the service ID and not the service system name
 * If specified system name does not exist then the API call would return no result/exception, which means we should take a decision of what to do if that happens. Maybe create an internal filter key that we know won't match any returned result? I'd prefer avoiding having special logic for this specific case (as in the future more filters would be demanded)

Pending tests until we know the approach of this PR.

@eguzki I'd like to receive your feedback on this. Thank you.


lib/3scale_toolbox/commands/activedocs_command/list_command.rb
```diff
@@ -42,8 +43,26 @@ def print_header(fields_to_show)
...
+          def filters
+            {
+              "service_id" => service_id_filter
+            }.compact
+          end
+
+          def filtered_activedocs(activedocs)
+            res = activedocs
+            filters.each do |filter_key, filter_value|
+              res = res.select{ |a| a.key?(filter_key) && a[filter_key].to_s == filter_value.to_s }
```
miguelsorianod: Convert values to strings to ensure comparisons can be made of the same type

eguzki:
I would implement filtering by service system_id, even though it required one api call more. Shouldn't be a problem for commands like this.

Regarding empty filter output, then let it be. Like when you request service activedocs and there are none. Only output headers are printed.

lib/3scale_toolbox/commands/activedocs_command/list_command.rb
```diff
@@ -42,8 +43,26 @@ def print_header(fields_to_show)
             puts fields_to_show.map { |e| e.upcase }.join("\t")
           end
 
+          def service_id_filter
+            options[:'service-id']
+          end
+
+          def filters
```
eguzki: Making extensible with a list of filters sounds good, but I think it is over engineering as I do not expect to have more filters to this command. Maybe yes? who knows.

However over engineering is nice sometimes and smart code is nice even if it is never extended. So I propose to implement an extensible filter list. There are different ways to do it. What I would do is to create a list of functions with the same interface, enumerable element as input param and filtered enumerable element as ouput. Each function would then be applied in a serial pipeline (one after the other connecting the output of the previous one with the input of the next one)

miguelsorianod:
> Regarding empty filter output, then let it be. Like when you request service activedocs and there are none. Only output headers are printed.

I meant that in case system_name does not exist then the filter would not make sense and ALL of the elements would be shown, not none

miguelsorianod:
> > Regarding empty filter output, then let it be. Like when you request service activedocs and there are none. Only output headers are printed.
> 
> I meant that in case system_name does not exist then the filter would not make sense and ALL of the elements would be shown, not none

Maybe set the filter to "-1" ? This assumes no service_id with id "-1" can exist in system

eguzki:
> Maybe set the filter to "-1" ? This assumes no service_id with id "-1" can exist in system

Can work. Other alternative is to replace the filter by service id for a '/dev/null' filter when service is not found. Anyway, warning about the service not found should be printed in stderr.

miguelsorianod:
@eguzki I tried to make the filters reusable across commands in case we need it. Can you review the new implementation and organization at class/module level?

Thank you.

lib/3scale_toolbox/commands/activedocs_command/list_command.rb
```diff
@@ -42,8 +43,32 @@ def print_header(fields_to_show)
...
+            res = []
+            if !service_ref_filter.nil?
+              res << AttributeFilters::ServiceIDFilterFromServiceRef.new(remote, service_ref_filter, "service_id")
+            end
+            res
+          end
+
+          def find_service(service_ref)
```
eguzki: This private method is being used?

lib/3scale_toolbox/commands/activedocs_command/list_command.rb
```diff
@@ -42,8 +43,32 @@ def print_header(fields_to_show)
...
+            Entities::Service::find(remote: remote, ref: service_ref)
+          end
+
+          def filtered_activedocs(activedocs)
```
eguzki: better use `Enumerable.reduce` method.

```
def filtered_activedocs(activedocs)
  filters.reduce(activedocs) { |current_list, filter| filter.call(current_list) }
end
```

lib/3scale_toolbox/attribute_filters/service_id_from_ref_filter.rb
```diff
@@ -0,0 +1,30 @@
+module ThreeScaleToolbox
+  module AttributeFilters
+    class ServiceIDFilterFromServiceRef
```
eguzki: What I would expect from class called `ServiceIDFilterFromServiceRef` is being initialized with some service_id value, instead of getting `remote` and service_ref.  The class also should know field name, service_id_key does not make sense IMO.

If service_id is nil, the filter should reject all values.

lib/3scale_toolbox/commands/activedocs_command/list_command.rb
miguelsorianod: :+1: that's great

lib/3scale_toolbox/attribute_filters/service_id_from_ref_filter.rb
miguelsorianod: > The class also should know field name, service_id_key does not make sense IMO.

The idea behind making the service_id_key configurable is that depending on the context the key name might change. For example, if using in a context where service is central (the service command for example) it's returned as "id" by System instead of by "service_id" :thinking:

miguelsorianod: > What I would expect from class called ServiceIDFilterFromServiceRef is being initialized with some service_id value, instead of getting remote and service_ref.

The idea behind this was to hide the logic of finding the service in the class itself so it would have not to be duplicated in every command code (client code of this class). Don't you see this approach suitable? You enter a service ref and you get the specified service refs from the list filtered

eguzki: Ok, let's go for it

eguzki:
LGTM, but still in WIP, I cannot approve

lib/3scale_toolbox/commands/activedocs_command/list_command.rb
miguelsorianod: If we are going to use the approach of finding the service in the Filter this does not belong here indeed anymore and it's not used. I'm going to remove it :+1:

miguelsorianod:
Tests and documentation added.
Ready for review @eguzki

PR #227, opened 2019-12-02
Do not set skip_swagger_validation when not provided as an option in activedocs apply

miguelsorianod:
Make activedocs apply command not set skip_swagger_validations on activedocs update.
Notice that this breaks compatibility with possible existing scripts so we should document in the new release that the scripts should be updated. It also changes default behavior, although it is a fix to a behavior that it should not exist.

### DIFF
--- a/spec/unit/commands/activedocs_command/apply_command_spec.rb
+++ b/spec/unit/commands/activedocs_command/apply_command_spec.rb
@@ -3,7 +3,8 @@ RSpec.describe ThreeScaleToolbox::Commands::ActiveDocsCommand::Apply::ApplySubco
   let(:activedocs) { instance_double(ThreeScaleToolbox::Entities::ActiveDocs) }
   let(:remote) { instance_double(ThreeScale::API::Client, 'remote') }
   let(:remote_name) { "myremote" }
-  let(:options) { {} }
+  let(:default_options) { { publish: false, hide: false } }
+  let(:options) { default_options }
   let(:activedocs_ref) { "activedocsref" }
   let(:activedocs_id) { "1" }
   let(:arguments) do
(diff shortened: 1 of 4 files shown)


miguelsorianod:
Ready for review @eguzki

spec/unit/commands/activedocs_command/apply_command_spec.rb
```diff
@@ -21,7 +22,7 @@
 
   context '#run' do
     context 'when --publish and --hide set' do
-      let(:options) { { publish: true, hide: true } }
+      let(:options) { default_options.merge({ publish: true, hide: true }) }
```
eguzki: Rubocop or any ruby linter does not complain for ` Redundant curly braces around a hash parameter`?

spec/unit/commands/activedocs_command/apply_command_spec.rb
```diff
@@ -87,16 +88,15 @@
           expect(activedocs_class).to receive(:find).with(remote: remote, ref: activedocs_ref).and_return(activedocs)
         end
 
-        context 'with no options' do
-          let(:options) { {} }
-
+        context 'with default options' do
           it 'activedocs not updated' do
+            expect(activedocs).not_to receive(:update)
```
eguzki: Not required.

If `expect` not set and activedocs receives `:update`, test will fail.

miguelsorianod: :+1:

spec/unit/commands/activedocs_command/apply_command_spec.rb
miguelsorianod: I was not aware of the need of usage of a linter in this project. I've fixed this and enabled rubocop in my IDE :+1:

miguelsorianod:
@eguzki ready for review

PR #229, opened 2019-12-05
Structured output: 

eguzki:
Added a new option:
```
    -o --output=<value>              Output format. One of: json|yaml
```
to the following commands:

- [x] method list, apply, create commands
- [x] application plans create, apply, list, show commands
- [x] metric list, apply, create commands
- [x] activedocs list, apply, create commands
- [x] application list, create, show, apply commands
- [x] service list, create, show, apply commands
- [x] proxy-config list, show commands

Other work done:

- [x] Unittests
- [x] Update doc

### DIFF
--- a/lib/3scale_toolbox/commands/application_command/apply_command.rb
+++ b/lib/3scale_toolbox/commands/application_command/apply_command.rb
@@ -206,12 +206,8 @@ module ThreeScaleToolbox
           end
 
           def printer
-            if options.key?(:output)
-              options.fetch(:output)
-            else
-              # keep backwards compatibility
-              CustomPrinter.new(resume: option_resume, suspend: option_suspend)
-            end
+            # keep backwards compatibility
+            options.fetch(:output, CustomPrinter.new(resume: option_resume, suspend: option_suspend))
           end
         end
       end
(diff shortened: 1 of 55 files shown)


eguzki:
Services and proxy configurations not implemented. Not sure they should be done as `services` and `proxy config` are deprecated entities in favor of `products` and `backends`

eguzki:
@miguelsorianod before doing tests and documentation, review would be good. Anything missing?

miguelsorianod:
It looks good to me

eguzki:
Iimplementing services and proxy configurations. Those concepts are being replaced by other names, but the 3scale API endpoints do not change, hence, command implementation will not change. Maybe command name, but not implementation.

eguzki:

lib/3scale_toolbox/commands/application_command/apply_command.rb
```diff
@@ -186,6 +204,15 @@ def application_ref
           def option_redirect_url
             options[:'redirect-url']
           end
+
+          def printer
```
miguelsorianod: Can we follow the same type of implementation that has been done in the printer method of lib/3scale_toolbox/commands/activedocs_command/list_command.rb (instead of if/else a options.fetch with a default fallback)?
If so, can we apply that type of implementation to all commands?

eguzki: :+1:

PR #232, opened 2020-01-13
Deprecate ruby 2.4

eguzki:
ruby 2.4 is almost in EOL date: 2020-03-31
~Support ruby 2.7~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,14 +27,6 @@ commands:
           command: |
             bundle exec rake spec:unit
 jobs:
-  ruby-246-unittest:
-    docker:
-      - image: circleci/ruby:2.4.6
-    working_directory: ~/repo
-    steps:
-      - checkout
-      - unittests
-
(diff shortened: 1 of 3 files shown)


thomasmaas:
system is switching to ruby 2.5 https://github.com/3scale/porta/pull/1521. Not better to keep the 2 in sync?

eguzki:
Since the toolbox is a gem, the idea is to support 2.5, 2.6 and 2.7. Hence the change:

```
spec.required_ruby_version = '>= 2.5'
```

eguzki:
Introducing support for ruby 2.7 comes with some issues as ruby 2.7 introduces bundler 2 and rubygem 3.  That is not backward compatible and unittests fails.

https://bundler.io/blog/2019/01/04/an-update-on-the-bundler-2-release.html

Enabling support for ruby 2.7 is not straightforward, so leaving for another PR when 2.7 is required.

PR #233, opened 2020-01-16
backend copy command

eguzki:
This command makes a copy of the referenced backend. Target backend will be searched by source backend system name. System name can be overriden with `--target_system_name` option.
If a backend with the selected `system_name` is not found, it will be created.        

Components of the backend being copied:
* metrics
* methods
* mapping rules

```
NAME
    copy - Copy backend

USAGE
    3scale backend copy [opts] -s <source_remote>
    -d <target_remote> <source_backend>

DESCRIPTION
    This command makes a copy of the referenced backend. Target backend will
    be searched by source backend system name. System name can be overriden
    with `--target_system_name` option. If a backend with the selected
    `system_name` is not found, it will be created.

    Components of the backend being copied:

    metrics

    methods

    mapping rules

OPTIONS
    -t --target_system_name=<value>      Target system name. Default to
                                         source system name
```

- [x] unittests
- [x] integration tests
- [x] Doc

PR #234, opened 2020-01-24
Copy service refactor

eguzki:
- [x] `3scale copy service` refactor to be reusable (by `3scale product copy`)
- [x] create new command `3scale service copy` running the same command as `3scale copy service`.
- [x] simply & clean copy service unittests and integration tests
- [x] relocate all `service copy` modules from `lib/3scale_toolbox/tasks` to `lib/3scale_toolbox/commands/service_command/copy_command`

Backend copy command PR #233  should be merged first


eguzki:

lib/3scale_toolbox/commands/service_command.rb
```diff
@@ -29,6 +30,7 @@ def self.command
       add_subcommand(Delete::DeleteSubcommand)
       add_subcommand(Create::CreateSubcommand)
       add_subcommand(Apply::ApplySubcommand)
+      add_subcommand(CopySubcommand)
```
miguelsorianod: This does not seem consistent with the other commands Where they have their own namespace. Might be a good idea to update it? not blocker but seems odd that the other have it and the copy one doesn't

eguzki: I think it makes sense to remove their own namespace. There is no use of it. Command tasks have their own namespace, but belongs to different directory as well. 

wdyt @miguelsorianod ?

miguelsorianod: Both approaches look good to me so up to you. In any case I think this would be a refactor that impacts all of them so we can do it in another PR/issue. To me this can be merged

PR #235, opened 2020-01-27
Product copy command

eguzki:
New command: `3scale product copy`
- [x] Command implementation
- [x] unittests
- [x] integration tests
- [x] Documentation

```
NAME
    copy - Copy product

USAGE
    3scale product copy [opts] -s <source_remote>
    -d <target_remote> <source_product>

DESCRIPTION
    This command makes a copy of the referenced product. Target product will
    be searched by source product system name. System name can be overriden
    with `--target_system_name` option. If a product with the selected
    `system_name` is not found, it will be created.

    Components of the product being copied:

    product configuration

    product settings

    product methods&metrics

    product mapping rules

    product application plans & pricing rules & limits

    product application usage rules

    product policies

    product backends

    product activedocs
```


eguzki:

lib/3scale_toolbox/commands/product_command/copy_command.rb
```diff
@@ -0,0 +1,78 @@
...
+module ThreeScaleToolbox
+  module Commands
+    module ProductCommand
+      class CopySubcommand < Cri::CommandRunner
+        include ThreeScaleToolbox::Command
+
+        def self.command
+          Cri::Command.define do
+            name        'copy'
+            usage       'copy [opts] -s <source_remote> -d <target_remote> <source_product>'
```
miguelsorianod: Can we change dashes instead of underscores for keywords with multiple words as we try to do in other places?

docs/copy-product.md
```diff
@@ -0,0 +1,66 @@
+## Copy Product
+
+This command makes a copy of the referenced product.
+Target product will be searched by source product system name.
+System name can be overriden with `--target_system_name` option.
+If a product with the selected `system_name` is not found, it will be created.
+
+Components of the backend being copied:
```
miguelsorianod: Can we document/explain the copy behavior when the destination product exists? (merge behavior)

lib/3scale_toolbox/commands/product_command/copy_command.rb
```diff
@@ -0,0 +1,78 @@
...
+            param   :source_product
+
+            runner CopySubcommand
+          end
+        end
+
+        def run
+          tasks = []
+          tasks << ThreeScaleToolbox::Commands::ServiceCommand::CopyCommand::CreateOrUpdateTargetServiceTask.new(context)
+          tasks << CopyCommand::DeleteTargetBackendUsagesTask.new(context)
```
miguelsorianod: Why do we delete the target backend usages before copying the source ones?

lib/3scale_toolbox/entities/backend_usage.rb
```diff
@@ -0,0 +1,99 @@
+module ThreeScaleToolbox
+  module Entities
+    ##
+    # BackendUsage represents Product - Backend mapping entry
+    class BackendUsage
```
miguelsorianod: Maybe some unit tests for this class?

lib/3scale_toolbox/commands/product_command/copy_command.rb
eguzki: Because 3scal API only allows updating path of a backend usage. So, if the path is already in use, there is no way to update target backend (backend_api_id attr of the backend usage object).

Hence, delete first approach has been followed. Happy to hear any other alternative.

lib/3scale_toolbox/commands/product_command/copy_command.rb
eguzki: :+1:

lib/3scale_toolbox/entities/backend_usage.rb
eguzki: :+1:

docs/copy-product.md
eguzki: :+1:

eguzki:
all change requests applied. Ready for review @miguelsorianod

lib/3scale_toolbox/commands/product_command/copy_command.rb
miguelsorianod: I made the original question due to I think the following scenario might happen with the current appraoch and I'm not sure that's what we want: If the destination product already exists, it might have backend ussages associated to backends that only exist on the destination but not in the source, and those would also be deleted during the copy and therefore lost. What do you think?

An alternative would be:

* Leave untouched the existing backendusages on the destination that don't have a matching backend system name (taken into account that this copy command itself might introduce them) and delete and recreate the backendusages that have a match.

miguelsorianod:
@eguzki additions/changes look good :+1: . I left some feedback on the delete backendusage conversation that we were having

lib/3scale_toolbox/commands/product_command/copy_command.rb
eguzki: Existing backend usages in the destination product that do not conflict (by path) with origin product backend usages are left and not deleted.

Your approach would work but it would require to build mapping between source backends and target backends. To save building this mapping, the implemented approach is:

> Existing backend usages in the destination product that do not conflict (by path) with origin product backend usages are left and not deleted.

This way, all source backend usages can be created safely in target product. Besides, unrelated target backend usages are not deleted.

eguzki:

lib/3scale_toolbox/commands/product_command/copy_command.rb
miguelsorianod: :+1:

PR #244, opened 2020-03-20
Threescale 669: Export apicast configuration file

eguzki:
New `proxy-config` subcommand: `export`.

Export proxy configuration to be used as 3scale apicast configuration: https://issues.redhat.com/browse/THREESCALE-669

```
$3scale proxy-config export -h
NAME
    export - Export proxy configuration for the entire provider account

USAGE
    3scale proxy-config export <remote>

DESCRIPTION
    Export proxy configuration for the entire provider account

    Can be used as 3scale apicast configuration file

    
    https://github.com/3scale/apicast/blob/master/doc/parameters.md#threescale_config_file

OPTIONS
       --environment=<value>      Gateway environment. Must be 'sandbox' or
                                  'production' (default: sandbox)
    -o --output=<value>           Output format. One of: json|yaml
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
    * Create, Apply, List, Delete [Method](docs/method.md)
    * Create, Apply, List, Show, Delete [Service](docs/service.md)
    * Create, Apply, List, Delete [ActiveDocs](docs/activedocs.md)
-   * List, Show, Promote [Proxy Configuration](docs/proxy-config.md)
+   * List, Show, Promote, Export [Proxy Configuration](docs/proxy-config.md)
    * [Copy Policy Registry](docs/copy-policy-registry.md)
    * Create, Apply, List, Show, Delete, Suspend, Resume [Applications](docs/applications.md)
    * [Remotes](docs/remotes.md)
--- a/docs/proxy-config.md
+++ b/docs/proxy-config.md
@@ -3,6 +3,7 @@
 * [List Proxy Configurations](#list)
 * [Show Proxy Configuration](#show)
 * [Promote Proxy Configuration](#promote)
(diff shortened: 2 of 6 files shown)


eguzki:

eguzki:
This PR was opened 18 days ago with no review yet. Deadline 20 days. Otherwise will be merged automatically.

cc @miguelsorianod

miguelsorianod:
Hi @eguzki,

Sorry for the delay. I've taken a look at it.

Looks good to me at functionality level.

I want to talk about the "semantics" of the command.

For what I see the command exports the proxyconfig for all services of a provider (remote in this case).
I've noticed the other "proxy-config" commands are at service level.

Would it make sense to have an "export" command for a single service? In case the answer is affirmative:
  * Should we think about what name to give to both of them in case in the future we want to implement the service-level one? two different commands to export, at service level or at tenant level. Or maybe in the future add a service parameter as optional?
  * What would be the difference with the show command when the export is at service level instead of at provider level?

eguzki:
> Should we think about what name to give to both of them in case in the future we want to implement the service-level one? two different commands to export, at service level or at tenant level. Or maybe in the future add a service parameter as optional?

I think that optional service parameter can be a future enhancement if there is any request about it. The apicast configuration itself also has allows [service filtering](https://github.com/3scale/APIcast/blob/master/doc/parameters.md#apicast_services_list)

> What would be the difference with the show command when the export is at service level instead of at provider level?
Show command dumps 3scale proxy object entity. The export command is meant to convert it to apicast configuration object. Currently the differences are small, but critical for the apicast, as it will not work without the update process.

PR #245, opened 2020-03-23
Application plan import: input resource as source of truth

eguzki:
Application plan import command takes as source of truth the imported resource

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,9 +5,9 @@ commands:
     steps:
       - restore_cache:
           keys:
-            - v1-dependencies-{{ checksum "Gemfile.lock" }}
+            - v2-dependencies-{{ checksum "Gemfile.lock" }}
             # fallback to using the latest cache if no exact match is found
-            - v1-dependencies
+            - v2-dependencies
 
       - run:
           name: Install dependencies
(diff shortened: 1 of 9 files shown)


eguzki:

miguelsorianod:
Can you provide some context about why do we want the behavior changed? What's the motivation of it? Is there some issue or something where the motivation is explained?

If I understand correctly, we have changed the behavior, where:
* Before: When importing application plan, we only import the pricing rules and limits that exist in the import file that at the same time don't exist in the service where we want to import them. It's like sort of a merge behaviour, but we don't merge the ones existing in both sides.
* Now: When importing an application plan, if there are any existing pricing rules and limits in the service we want to import to, we delete all of them and we import the ones defined in the import file. It's sort of a set behaviour, where we set what is defined in the import file side

Am I correct?

Thank you.

eguzki:
> Can you provide some context about why do we want the behavior changed? What's the motivation of it? Is there some issue or something where the motivation is explained?

https://issues.redhat.com/browse/THREESCALE-3874
https://issues.redhat.com/browse/THREESCALE-3211

>* Before: When importing application plan, we only import the pricing rules and limits that exist in the import file that at the same time don't exist in the service where we want to import them. It's like sort of a merge behaviour, but we don't merge the ones existing in both sides.
> * Now: When importing an application plan, if there are any existing pricing rules and limits in the service we want to import to, we delete all of them and we import the ones defined in the import file. It's sort of a set behaviour, where we set what is defined in the import file side
Am I correct?

Yes your are correct. The main implication with the change in the "merge strategy" is that with the *SET* semantics, existing limits and pricing rules not related with the imported file content will be deleted as well.

PR #246, opened 2020-03-23
Standard way to report errors

### LINKED ISSUE #80: RFE: Have a standard way to report errors
The 3scale toolbox is being improved to become the recommended way to configure and deploy services in 3scale. This fits in a broader approach of Continuous Delivery that has been started by several customers / communities (I personally maintain [the threescale-cicd ansible role](https://github.com/nmasse-itix/threescale-cicd). Those approaches are using different means to drive the 3scale admin portal, some are using the 3scale-cli other (like the threescale-cicd role) are using the 3scale REST APIs. 

If we want to make the 3scale_toolbox the de facto standard for interacting with 3scale, we need it to be usable in a CI/CD approach. 

At a glance, here are the features I expect from the 3scale toolbox :

- #81: Is idempotent
- #82: Works with system_name and not service_id
- #83: Can output human readable information as well as json or yaml (akin to `oc cmd -o yaml`)
- #80: Have a standard way to report errors

## A standard way to report errors

Since the 3scale toolbox will be used from scripts / pipelines / etc, any error must be reported without any ambiguity. 

### Proposition

The plugins have a standard way to report errors, the unix return code is set appropriately and the output can be parsed easily by a script (JSON / YAML for instance). 

### Examples

If the requested operation cannot be completed, the Unix return code is > 1 and if the operation has been completed successfully the Unix return code is 0.
If the requested operation cannot be completed, the output would look like :

```json
{
  "error_code": "E_CANNOT_READ_OPENAPI",
  "error_message": "Cannot parse OpenAPI Specs: missing field bla bla bla",
  "kind": "local",
  "stacktrace": "..."
}
```

This feature needs to be in the toolbox's core so that every plugin can leverage this feature.

eguzki:
Fixes #80

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -1,0 +1,67 @@
+## Error Reporting
+
+Since the 3scale toolbox is used from scripts / pipelines / etc, any error must be reported without any ambiguity.
+
+## Table of contents
+* [Unix Return Code](#unix-return-code)
+* [Error Report Fields](#error-report-fields)
+* [Error Types](#error-types)
+* [Error Code Table](#error-code-table)
+
+### Unix Return Code
+
+If the requested operation cannot be completed, the Unix return code is *1* and if the operation has been completed successfully the Unix return code is *0*.
+
+### Error Report Fields
+
+When an error occurs, the toolbox writes to *stderr* an structured error report with some fields.
+
+```
+{
+  "code": string,
+  "message": string,
+  "class": string,
+  "stacktrace": array of string
+}
+```
+
+| Field | Description |
+| --- | --- |
+| code | Error code. See [table](#error-code-table) |
+| message | Description text of the error |
+| class | Class of the error |
+| stacktrace | (optional) the stacktrace causing the error |
+
+### Error Types
+
+There are three types of errors and the report will depend on the type.
+
+#### Managed Error
+
+The error is managed by the toolbox. This means the error is created by the toolbox in a controlled way.
+
+#### Unmanaged Error
+
+The error is not managed by the toolbox. But still, the error is not a *panic* error.
+The application has been able to handled it and generate detailed report for the user to inspect.
+For example, `RuntimeError` or `ZeroDivisionError` are included in this class of errors.
+
+**For this type of errors, the toolbox generates a detailed report in `crash.log` file in the current working directory.**
+
+The assigned error code for this type of errors is `E_UNKNOWN`. See [table](#error-code-table)
+
+#### Unhandled Error
+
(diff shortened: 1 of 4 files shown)


eguzki:

docs/errors.md
```diff
@@ -0,0 +1,67 @@
...
+The application has been able to handled it and generate detailed report for the user to inspect.
+For example, `RuntimeError` or `ZeroDivisionError` are included in this class of errors.
+
+**For this type of errors, the toolbox generates a detailed report in `crash.log` file in the current working directory.**
+
+The assigned error code for this type of errors is `E_UNKNOWN`. See [table](#error-code-table)
+
+#### Unhandled Error
+
+This type of errors are not handled by the toolbox and the error report is delegated to the ruby framework, usually showing stacktrace and error message.
```
miguelsorianod, line 55: Does this mean that in this case no 3scale toolbox json output is produced?

docs/errors.md
```diff
@@ -0,0 +1,67 @@
...
+  "class": string,
+  "stacktrace": array of string
+}
+```
+
+| Field | Description |
+| --- | --- |
+| code | Error code. See [table](#error-code-table) |
+| message | Description text of the error |
+| class | Class of the error |
```
miguelsorianod, line 32: Maybe specify that class is a Ruby class?
Before looking at the code I thought it could be a Ruby class or the "class" of the error (conceptually)

docs/errors.md
miguelsorianod: If 3scale toolbox json output is produced, What would be the code in the code table?

miguelsorianod:
As this changes the output content (now that I think about it, the same happens with json structured output), should we add some info in the future release notes telling that the shown output of some commands has changed? (in case someone is relying on that)

miguelsorianod:
Looks good to me.

Just some minor comments about how it works and feedback

docs/errors.md
eguzki: In this case no 3scale json output is produced, deliberately

docs/errors.md
eguzki: IMO user does not need to know that.

eguzki:
> As this changes the output content (now that I think about it, the same happens with json structured output), should we add some info in the future release notes telling that the shown output of some commands has changed? (in case someone is relying on that)

The json structured output does not break backward compatibility, as it is optional. 

Yes, indeed, when there are errors, this PR changes the output (in stderr). Adding release notes about this is a good idea

PR #250, opened 2020-04-17
service copy command fix: do not copy endpoint url hosted services

eguzki:
Service copy command: do not copy endpoint url for services with hosted gateway

### DIFF
--- a/lib/3scale_toolbox/commands/service_command/copy_command/copy_service_proxy_task.rb
+++ b/lib/3scale_toolbox/commands/service_command/copy_command/copy_service_proxy_task.rb
@@ -6,9 +6,25 @@ module ThreeScaleToolbox
           include Task
 
           def call
-            target.update_proxy source.proxy
+            target.update_proxy target_proxy_attrs
             target.update_oidc source.oidc if source.attrs['backend_version'] == 'oidc'
             puts "updated proxy of #{target.id} to match the original"
+          end
+
+          def target_proxy_attrs
+            if source.attrs['deployment_option'] == 'hosted'
+              # for services with "hosted" deployment config,
+              # apicast gateway addresses should not be copied.
+              # Gateway addresses includes service's id and if copied, the service will not
+              # have visibility.
+              source_proxy.dup.delete_if { |key, _v| %w[endpoint sandbox_endpoint].include? key }
+            else
+              source_proxy
+            end
+          end
+
+          def source_proxy
+            @source_proxy ||= source.proxy
           end
         end
       end
--- a/lib/3scale_toolbox/commands/service_command/copy_command/copy_service_proxy_task.rb
+++ b/lib/3scale_toolbox/commands/service_command/copy_command/copy_service_proxy_task.rb
@@ -13,10 +13,9 @@ module ThreeScaleToolbox
 
           def target_proxy_attrs
             if source.attrs['deployment_option'] == 'hosted'
-              # for services with "hosted" deployment config,
-              # apicast gateway addresses should not be copied.
-              # Gateway addresses includes service's id and if copied, the service will not
-              # have visibility.
+              # For services with "hosted" deployment config, 
+              # "Public Base URL" should not be copied, mainly because public base URL is self-assigned.
+              # Two 3scale products (aka services) cannot be served using the same public base URL.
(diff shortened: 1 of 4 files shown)


lib/3scale_toolbox/commands/service_command/copy_command/copy_service_proxy_task.rb
```diff
@@ -6,10 +6,26 @@ class CopyServiceProxyTask
...
             puts "updated proxy of #{target.id} to match the original"
           end
+
+          def target_proxy_attrs
+            if source.attrs['deployment_option'] == 'hosted'
+              # for services with "hosted" deployment config,
+              # apicast gateway addresses should not be copied.
+              # Gateway addresses includes service's id and if copied, the service will not
+              # have visibility.
+              source_proxy.dup.delete_if { |key, _v| %w[endpoint sandbox_endpoint].include? key }
```
miguelsorianod, line 20: I'm trying to understand the motivation of this.

What happens when the endpoint is not copied?Is a new one autogenerated in that case?

Why would it not be visible? By "Gateway addresses" you mean the url?

Reading at the 3scale api docs documentation, the "Proxy Update" rest method provides a field "endpoint", which is documented as "Public Base URL for production environment". Why this can't be an arbitrary value? Should it be documented there that those fields should only be used when you use a hosted apicast gateway?

eguzki: "Public Base URL for production environment" can be arbitrary value for self-managed deployments. And even then, cannot be arbitrary. Two 3scale products (aka services) cannot be served using the same public base URL. The other way around is not possible either, if I am not wrong. One 3scale product cannot be served from two different public base URL's. 

For *hosted* deployments, an example can illustrate the issue. Let's assume public base URL is copied.

There is a Product A with public URL `a.staging.example.com`. Your API uses application A1, so your requests would look like: `GET a.staging.example.com/?user_key=A1`.

Then you copy product A to some product named A'. The public URL will be `a.staging.example.com` (because it has been copied). We create a new application A2 for the product A'. We try the following request:

`GET a.staging.example.com/?user_key=A2`

The result will be `HTTP/1.1 403 Forbidden`. Why? because the public base URL identifies the product, which is A (not A') and application A2 belongs to A', hence does not exist in product A. Thus, Forbidden.

eguzki: I am going to improve the comment in the code as it is not clear enough.

eguzki: Question: does copying public URL make sense for self managed deployments? @miguelsorianod

miguelsorianod: The explanation about hosted deployments make sense in the copy context yes :+1:

miguelsorianod: > I am going to improve the comment in the code as it is not clear enough.

:+1: if you think it can be improved, but what I was really wondering is if the ActiveDocs (not the toolbox code comment) should be improved regarding that.

miguelsorianod: > Question: does copying public URL make sense for self managed deployments? @miguelsorianod

I think this totally depends on what is to be used for.
For example, if you want to duplicate the environment because you want to give the new environment another use, and the product public url endpoint would be a public DNS it might not make sense to copy it, because a public DNS name is assigned to a unique resolving IP (excluding load balancing scenarios of course) and you would have a resolving problem.

But with the same dns context as before, if you want for example to duplicate the environment because you are migrating it and you'll delete the old one then it might make sense.

eguzki: For products configured with hosted gateways, does not make sense to copy public base URL in any case, because public base URL is self-assigned. 

So I will make it conditional and document it.

* Hosted deployments -> do not copy public base URL
* Self managed deployments -> copy public base URL

eguzki:
Doc updated and improved source code comments for clearness.

PR #252, opened 2020-04-24
fix cleanup integration tests

eguzki:


eguzki:

PR #260, opened 2020-07-01
Support ruby2.7

eguzki:


Gemfile.lock
```diff
@@ -109,6 +109,3 @@ DEPENDENCIES
   rspec (~> 3.8)
   rubyzip (>= 1.3.0)
   webmock (~> 3.4)
-
-BUNDLED WITH
```
eguzki, line 113: wdty @davidor ? Is it safe to remove "BUNDLED WITH" from Gemfile.lock?

ruby 2.7 does not support Bundle 1.17, so dependencies could not be met and tests failed.

Removing this, tests passes for supported ruby versions: 2.5, 2.6 and 2.7

davidor: I think this just alerts you if you try to use an older version of bundler than the one specified.

Re-generating this file with a newer version of bundler supported by Ruby 2.7 could be better than simply removing these 2 lines.

eguzki: I am afraid I cannot add 
```
BUNDLED WITH
   2.1.4
```

ruby 2.5 and ruby 2.6 tests fails as dependencies cannot be met.

```
#!/bin/bash -eo pipefail 
bundle install --jobs=3 --retry=3 --path vendor/bundle 
 

Traceback (most recent call last):
 
	2: from /usr/local/bin/bundle:23:in `<main>'
 
	1: from /usr/local/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
 
/usr/local/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.1.4) required by your /home/circleci/repo/Gemfile.lock. (Gem::GemNotFoundException)
 
To update to the latest version installed on your system, run `bundle update --bundler`.
 
To install the missing version, run `gem install bundler:2.1.4`
 
 
Exited with code exit status 1 

CircleCI received exit code 1 
```

eguzki: Bundle version will be changed in the environment for ruby 2.5 and 2.6 as mentioned [here](https://dev.to/tanakaworld/how-to-resolve-you-must-use-bundler-2-or-greater-with-this-lockfile-2pf7)

miguelsorianod (approved):

PR #275, opened 2020-12-03
Change base image: ruby:2.7

eguzki:
Currently, `centos/ruby-25-centos7:latest` base image with tag `centos/ruby-25-centos7@sha256:71a4818652959d2c03e98b42ad74571426fb5ccab3f93448ba1dd2524a6ad0ef`  breaks the build:

```
$ docker run --rm -ti quay.io/redhat/3scale-toolbox:master 3scale help
/usr/bin/container-entrypoint: line 2: exec: 3scale: not found
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,15 @@
-FROM centos/ruby-25-centos7:latest
+FROM ruby:2.7
 MAINTAINER Eguzki Astiz Lezaun <eastizle@redhat.com>
 
-USER root
-WORKDIR /opt/app-root/src
+WORKDIR /usr/src/app
 COPY . .
-RUN /bin/bash -l -c "gem build 3scale_toolbox.gemspec"
-RUN /bin/bash -l -c "gem install 3scale_toolbox-*.gem --no-document"
+RUN gem build 3scale_toolbox.gemspec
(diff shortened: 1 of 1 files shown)


gsaslis:
@eguzki I have also been noticing some failures in downstream jobs that must be related to changes in RHSCL base images (something related with env vars and paths I think), but haven't found the time to dig further. Any chance you already know why ?

miguelsorianod (approved):

eguzki:
> @eguzki I have also been noticing some failures in downstream jobs that must be related to changes in RHSCL base images (something related with env vars and paths I think), but haven't found the time to dig further. Any chance you already know why ?

I am so curious about that. I know it has to do with RHSCL base installation but the rabbit hole can be deep or just a little issue. I am more afraid of wasting my time in the rabbit hole than curious. Hence, I changed the base image for the standard one.

gsaslis:
@eguzki haha - exact same situation here! :D 

...but I might be forced to look into that anyway - I'll let you know if i find out why.  ;)

miguelsorianod:
Having changed the base image here, the change should be also applied to productised version of toolbox because there's been a change of ruby version

eguzki:
The productised version does not have to be changed. We can maintain both, does take a lot of effort.

PR #281, opened 2021-01-25
new command: 3scale proxy-config deploy

eguzki:
Command to promote the APIcast configuration to the Staging Environment 

Only supported in 3scale release >= 2.7

Requires: https://github.com/3scale/3scale-api-ruby/pull/94

Jira Issue: https://issues.redhat.com/browse/THREESCALE-6620

### DIFF
--- a/docs/proxy-config.md
+++ b/docs/proxy-config.md
@@ -4,6 +4,7 @@
 * [Show Proxy Configuration](#show)
 * [Promote Proxy Configuration](#promote)
 * [Export Proxy Configuration](#export)
+* [Deploy Proxy Configuration](#deploy)
 
 ### List
 
@@ -103,5 +104,22 @@ DESCRIPTION
 OPTIONS
        --environment=<value>      Gateway environment. Must be 'sandbox' or
                                   'production' (default: sandbox)
+    -o --output=<value>           Output format. One of: json|yaml
+```
+
+### Deploy
+
+```shell
+NAME
+    deploy - Promotes the APIcast configuration to the Staging Environment
+
+USAGE
+    3scale proxy-config deploy <remote> <service>
+
+DESCRIPTION
+    Promotes the APIcast configuration to the Staging Environment (Production
+    Environment in case of Service Mesh).
+
+OPTIONS
     -o --output=<value>           Output format. One of: json|yaml
 ```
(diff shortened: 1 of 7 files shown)


eguzki:

docs/proxy-config.md
```diff
@@ -105,3 +106,20 @@ OPTIONS
                                   'production' (default: sandbox)
     -o --output=<value>           Output format. One of: json|yaml
 ```
+
+### Deploy
+
+```shell
+NAME
+    deploy - Promotes the APIcast configuration to the Staging Environment
```
miguelsorianod, line 114: Why has it been done as a standalone command and not part of the `promote` command?

docs/proxy-config.md
```diff
@@ -105,3 +106,20 @@ OPTIONS
...
+
+USAGE
+    3scale proxy-config deploy <remote> <service>
+
+DESCRIPTION
+    Promotes the APIcast configuration to the Staging Environment (Production
+    Environment in case of Service Mesh).
```
miguelsorianod, line 121: What's the particularity of service mesh?

miguelsorianod (commented):
Looks good to me. I've left some questions

docs/proxy-config.md
eguzki: Copied from 3scale API docs. TBH I do not know. Maybe in ServiceMesh there is no staging env

docs/proxy-config.md
eguzki: promote command, as it says in its description, promotes latest staging conf to production env
```
3scale proxy-config promote -h
NAME
    promote - Promote latest staging Proxy Configuration to the production environment
```

miguelsorianod: That's the current description.

Why has not been placed for example as a flag of the promote command, for example `--environment <env>`?

eguzki: It would imply full refactoring of the command, and still will not be very clear. For instance, check `promotable?` method being used by the current command. That will not make sense. 

I think it is cleaner to have its own command instead of being part of another command using optional input param

miguelsorianod: :+1:

miguelsorianod (approved):

PR #284, opened 2021-02-01
Fix proxy config integration tests

eguzki:
Proxy config integration tests are failing 80% of the runs. 

The main reason: Sometimes (more often than expected) `update_proxy` operation does not deploy proxy configuration to "staging" or sandbox. It is not eventually deployed, it is never deployed. Used `proxy deploy` operation to actually deploy proxy configuration to staging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,7 +145,7 @@ workflows:
           type: approval # <<< This key-value pair will set your workflow to a status of "On Hold"
           # On approval of the `hold` job, any successive job that requires the `hold` job will run.
       - integration_tests:
-          context: supertestaccount
+          context: autotestaccount
           requires:
             - manual_approval
   nightly:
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -145,7 +145,7 @@ workflows:
           type: approval # <<< This key-value pair will set your workflow to a status of "On Hold"
           # On approval of the `hold` job, any successive job that requires the `hold` job will run.
       - integration_tests:
-          context: supertestaccount
+          context: autotestaccount
```
miguelsorianod, line 148: I guess the context has been renamed or a new one is being used now for some reason?

eguzki: autotestaccount provider account used to be old 3scale Service based. It has been "promoted" to APIaP (products and Backends), so e2e tests can be run in this env. I use "supertest" for daily use, so I try to avoid conflicts.

miguelsorianod (approved):

.circleci/config.yml
miguelsorianod: :+1

PR #285, opened 2021-02-08
Refactoring and general improvements

eguzki:
* Cache remote metrics methods
* Service and backend methods only having hits as parent
* Mapping rules entity
* Limit entity
* PricingRule entity
* Service returns metrics, methods and mapping rules as entities (like backend does)


lib/3scale_toolbox/remote_cache.rb
```diff
@@ -0,0 +1,153 @@
...
+      return methods_cache[key] if methods_cache.has_key? key
+
+      subject.list_methods(service_id, metric_id).tap do |methods|
+        methods_cache[key] = methods unless methods.respond_to?(:has_key?) && !methods['errors'].nil?
+      end
+    end
+
+    def create_metric(service_id, attributes)
+      subject.create_metric(service_id, attributes).tap do |_|
+        metrics_cache.delete(service_id)
```
miguelsorianod, line 37: When creating a metric, why is the metric being deleted from the cache and not overwritten it?
What happens with the possible error result on `create_metric`?

lib/3scale_toolbox/remote_cache.rb
```diff
@@ -0,0 +1,153 @@
...
+      end
+    end
+
+    def update_metric(service_id, metric_id, attributes)
+      subject.update_metric(service_id, metric_id, attributes).tap do |_|
```
miguelsorianod, line 42: Do we ignore errors on the possible call to update_metric?
This question applies to other similar methods

lib/3scale_toolbox/remote_cache.rb
eguzki: When the existing cache is being invalidated because a metric is deleted or updated, the naive implementation is to delete the existing cache. Smarter implementation is postponed a good nice to have. But better in another PR.

Implemented error handling on create and update methods

lib/3scale_toolbox/remote_cache.rb
eguzki: answered

lib/3scale_toolbox/remote_cache.rb
miguelsorianod: I understand about invalidating the cache when the metric is being deleted or updated and it makes total sense, but this specific case is about metric creation :thinking:

eguzki: The same for metric update, you can update your cached data with the updated metric. Can  be easily done. But not sure we want to enter into  this issue. Not for this PR. IF we implement caching update, the cache will never be invalidated for the entire command lifetime. I am not sure we should implement.

miguelsorianod: :+1:

spec/unit/remote_cache_spec.rb
```diff
@@ -0,0 +1,457 @@
...
+  let(:error_response) { {'errors' => {}} }
+  subject { described_class.new(proxied_object) }
+
+  it 'list_metrics first call cache miss' do
+    expect(proxied_object).to receive(:list_metrics).and_return(metrics)
+    expect(subject.list_metrics(service_id)).to eq(metrics)
+  end
+
+  it 'list_metrics second call cache hit' do
+    expect(proxied_object).to receive(:list_metrics).and_return(metrics)
```
miguelsorianod, line 16: Shouldn't this to receive "once" ?

miguelsorianod (approved):
I left a minor comment

spec/unit/remote_cache_spec.rb
eguzki: it is expecting to receive one time only. The second call would fail. Unless you change to something like this:
```
expect(proxied_object).to receive(:list_metrics).twice.and_return(metrics)
```

or

```
allow(proxied_object).to receive(:list_metrics).and_return(metrics)
```

PR #286, opened 2021-02-08
Product import export command

eguzki:
Two new commands
* `3scale product export`: export 3scale product and used backends in [Product CRD](https://github.com/3scale/3scale-operator/blob/3scale-2.10.0-CR2/doc/product-reference.md) format and [Backend CRD](https://github.com/3scale/3scale-operator/blob/3scale-2.10.0-CR2/doc/backend-reference.md) format
* `3scale product import`: import 3scale product and used backends in [Product CRD](https://github.com/3scale/3scale-operator/blob/3scale-2.10.0-CR2/doc/product-reference.md) format and [Backend CRD](https://github.com/3scale/3scale-operator/blob/3scale-2.10.0-CR2/doc/backend-reference.md) format

Requires enhancements from #285


lib/3scale_toolbox/commands/product_command/import_command.rb
```diff
@@ -0,0 +1,125 @@
...
+
+        def run
+          validate_artifacts_resource!
+
+          product_list.each do |product|
+            context = {
+                target_remote: remote,
+                source_remote: crd_remote,
+                source_service_ref: product.system_name,
+                logger: Logger.new('/dev/null')
```
miguelsorianod, line 31: Is this intended to be /dev/null?

lib/3scale_toolbox/crds/application_plan_dump.rb
```diff
@@ -0,0 +1,19 @@
+module ThreeScaleToolbox
+  module CRD
+    module ApplicationPlan
```
miguelsorianod, line 3: Should this be called  `ApplicationPlanSerializer` to follow naming convention?

lib/3scale_toolbox/crds/backend_parser.rb
```diff
@@ -0,0 +1,55 @@
+module ThreeScaleToolbox
+  module CRD
+    class BackendParser
+      Metric = Struct.new(:system_name, :friendly_name, :description, :unit)
```
miguelsorianod, line 4: Here we are defining our own custom structs.

There are their corresponding 'Entity' classes, can't we reuse them? some classes on the model will be using this custom structs and others will be using the entities one, which is kind of confusing.

lib/3scale_toolbox/commands/product_command/import_command.rb
eguzki: Yes, for backwards compatibility, the `product copy` command still prints messages. But for the new command, we do not want those messages to be printed. Instead, the new command generates in stdout a report in json format with structured info.

lib/3scale_toolbox/crds/backend_parser.rb
eguzki: This `Metric` struct is private for the `BackendParser` and should not be used externally. 

The entities are interface to external remote objects and have methods and required input parameters (like remote) that do not make sense for this structs. They only share the list of attributes.

lib/3scale_toolbox/crds/application_plan_dump.rb
eguzki: done

lib/3scale_toolbox/commands/product_command/import_command.rb
miguelsorianod: :+1:

lib/3scale_toolbox/crds/backend_parser.rb
miguelsorianod: :+1:

miguelsorianod (approved):

PR #308, opened 2021-06-23
[breaking change] mapping rules will be replaced instead of adding missing rules

### LINKED ISSUE #306: Old mapping rules aren't replaced when importing a product
When importing a pre-existing product with different mapping rules, I expect the new product definition to completely replace the old one. Instead, the new mapping rules are added to the old ones. We are left with a product that contains both the old unwanted mapping rules and the new mapping rules. If gone unnoticed, this is a potential security risk.

eguzki:
https://issues.redhat.com/browse/THREESCALE-7215

Fixes #306

### DIFF
--- a/docs/copy-backend.md
+++ b/docs/copy-backend.md
@@ -11,7 +11,8 @@ Components of the backend being copied:
 * methods
 * mapping rules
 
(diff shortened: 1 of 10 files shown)


eguzki:

miguelsorianod (approved):

PR #309, opened 2021-06-26
docker image ubi8/ruby-27

eguzki:


miguelsorianod:
What's the reason of this change?

eguzki:
ubi8 supported by RH and far away End Of Life

miguelsorianod:
> ubi8 supported by RH and far away End Of Life

I'm not entirely sure that we want to do this:

* Do we want to use the ubi one by RH in the toolbox community side? 
* `registry.redhat.io registry also requires users having an account. This means if they want to build it they will need to have an account. We are adding an additional needed step to build it on the community side.

eguzki:
>  if they want to build it

I though about that too, but IMO if they want to build it, then those have developer profile as they need Docker knowledge. Thus, it makes kind of sense to ask for RH developer account.

miguelsorianod:
> I thought about that too, but IMO if they want to build it, then those have developer profile as they need Docker knowledge. Thus, it makes kind of sense to ask for RH developer account.

Even if they are developers I don't see a need of requiring a RH developer account.
Related to this, I've just noticed that in the red hat  catalog container explorer: https://catalog.redhat.com/software/containers/ubi8/ruby-27/5f521b90acb94fdd904e98d7?container-tabs=gti
the URL being returned now is `registry.access.redhat.com` instead of `registry.redhat.io` and it seems it does not require an account. This seems it might be recent change. Maybe we should change to that in community (and in product still use registry.redhat.io) to not require the account?

eguzki:
The `registry.access.redhat.com` domain is a persistent one? it is not temporary? I thought `registry.access.redhat.com` was going to be deprecated. If not, we can, indeed, change. Good point.

miguelsorianod:
> The `registry.access.redhat.com` domain is a persistent one? it is not temporary? I thought `registry.access.redhat.com` was going to be deprecated. If not, we can, indeed, change. Good point.

That was my understanding too but I noticed that `registry.access.redhat.com` is now shown whereas I seem to remember before `registry.redhat.io` was being shown (I might be wrong) so it seems it might be either a change in behavior or an error on the catalog explorer

miguelsorianod:
In any case, I reiterate that I'm not sure this change is needed at all. I don't see a need to use a redhat-specific image in the community side of toolbox

eguzki:
It is a matter of converging both images and if the base image can be respined to include new CVE fixes, the better. I think it is better that rely on `ruby:27`

miguelsorianod:
`ruby:27` is also updated and I think it includes cve fixes, with a community-effort basis.
I understand your point of converging both images. It seems a tradeoff between converging the image and increasing level of entry to build/develop 3scale toolbox

miguelsorianod (approved):

PR #310, opened 2021-06-29
bundler using v2.2.21

eguzki:
Fix for CVE-2020-36327 rubygem-bundler: Dependencies of gems with explicit source may be installed from a different source

https://issues.redhat.com/browse/THREESCALE-7085


miguelsorianod:
What is the motivation of this change?

eguzki:
> What is the motivation of this change?

Added in the description

miguelsorianod (approved):
:+1:

PR #311, opened 2021-07-01
Fix circleci main branch

eguzki:


.circleci/config.yml
```diff
@@ -120,7 +120,7 @@ jobs:
       - run:
           name: Docker container help test
           command: |
-            docker run --rm -t quay.io/redhat/3scale-toolbox:master 3scale help
+            docker run --rm -t quay.io/redhat/3scale-toolbox:main 3scale help
```
miguelsorianod, line 123: Is the quay.io repository appropriately dealing with creating the `main` image now instead of `master`?
At the moment I don't see any `main` tag

eguzki: WIP. There is an issue with the build triggers in quay. I will let you know as the nightly test (` docker_container_help_test`) will fail if quay builds are not working.

eguzki: `quay.io/redhat/3scale-toolbox:main` has been created. 

It is safe to merge this

miguelsorianod (approved):

PR #314, opened 2021-07-07
Gemfile.lock: add powerpc64le-linux platform

eguzki:
Add powerpc64le-linux platform

The outcome of executing
```
bundle lock --add-platform powerpc64le-linux
```

Bundler is smart enough to use platform dependent gems whenever available in the lock file. This is desired as nokogiri [recommends](https://github.com/sparklemotion/nokogiri#native-gems-faster-more-reliable-installation) using native gems with precompiled binaries

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,12 +46,16 @@ GEM
       with_env (= 1.1.0)
       xml-simple
     method_source (1.0.0)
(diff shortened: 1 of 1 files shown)


gsaslis (approved):
Great to see ppc64le support being added here! 👍

PR #316, opened 2021-08-03
remove warning from ruby 2.7 when last arg is keyword

eguzki:


miguelsorianod:
It seems the build has errored

eguzki:
> It seems the build has errored

That is travisCI with powerPC build. It fails frequently. I have pending in my TODO list to check that. However, required tests to pass are those from the circleCI and they all passed. 

I have rebased, as master head changed, and let's see the test results

eguzki:
CircleCI tests passed. Only approval is missing

miguelsorianod (approved):

PR #327, opened 2021-09-09
enhance importing OpenAPI doc

eguzki:
Explained how 3scale is configured from the OpenAPI document. Which parts of the OpenAPI document are used, and how, to configure 3scale.

Also added documentation about the minimum OpenAPI Document requried to configure a working 3scale product.

https://issues.redhat.com/browse/THREESCALE-4712

### DIFF
--- a/docs/openapi.md
+++ b/docs/openapi.md
@@ -1,36 +1,225 @@
 ## Import API definition to 3scale from OpenAPI definition
 
-Features:
+To create a new service or to update an existing service, you can import the OpenAPI definition from a local file or a URL.
 
-* OpenAPI __2.0__ specification (f.k.a. __swagger__)
-* OpenAPI __3.0.2__ specification
-  * Limitations:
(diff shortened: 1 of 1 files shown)


docs/openapi.md
```diff
@@ -1,36 +1,251 @@
...
+```
+
+When OpenAPI does not specify any security requirements:
+* The product is considered as an "Open API"
+* `default_credentials` 3scale policy will be added (also called as `anonymous_policy`)
+* The command option `--default-credentials-userkey` is required and the command will fail if not provided.
+
+#### ActiveDocs
+
+A 3scale ActiveDoc will be created (or updated if previously existed).
```
miguelsorianod, line 169: Does that mean that if it previously existed it can alter it in some way? is this controllable? what can potentially differ?

miguelsorianod (approved):
Looks good to me :+1:  I just left a minor question

docs/openapi.md
eguzki: For example, when you add/delete/update an operation, the activedoc object is updated accordingly.

miguelsorianod: :+1:

PR #331, opened 2021-09-17
import/export application plans support 3scale backends

eguzki:
The jira issue: https://issues.redhat.com/browse/THREESCALE-7568 revealed that import/export application plan commands do not support 3scale backends.

This PR adds support for exporting/importing application plans with limits and pricing rules on backend metrics. 

Some refactor was done to implement the approach of product import/export commands. Basically, each 3scale entity class has `to_hash` method that implements the serialization logic.

Added validation check to the `import` command that was missing. 

TODO:
- [X] Application plan export command
- [x] Application plan import command
- [x] Adapt unittests
- [x] Adapt integration tests to include limits and pricing rules on backend metrics


lib/3scale_toolbox/commands/plans_command/import/step.rb
```diff
@@ -82,8 +98,14 @@ def find_feature_by_system_name(system_name)
             service_features.find { |feature| feature['system_name'] == system_name }
           end
 
-          def find_metric_by_system_name(system_name)
-            service_metrics_and_methods.find { |metric| metric.system_name == system_name }
+          def find_metric(metric_system_name, backend_system_name)
```
miguelsorianod, line 101: The name should probably be `metric_backend_system_name`

miguelsorianod, line 101: The idea of this method is that only one of the two can be nil right?

eguzki: the callers of this ruby method have
* metric system_name
* (optionally) backend system name

They want to get the actual metric object out of it. Also, this method is usefull to check if the metric reference is valid (the metric exists or not)

It is not always a backend's metric. That's why I think that is not a good name. But I am not good naming methods. Propose an alternative :)

eguzki: No, metric_system_name must always be non-nil. backend_system_name can be nil. Maybe we can have it with a default value to make it explicit.

```
          def find_metric(metric_system_name, backend_system_name = nil)
```

spec/unit/commands/plans_command/import/validation_plan_step_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+  let(:backend_usage_02) { instance_double(ThreeScaleToolbox::Entities::BackendUsage, 'backend_usage_02') }
+  let(:backend_usage_list) { [backend_usage_01, backend_usage_02] }
+  let(:context) do
+    {
+      service: service,
+      artifacts_resource: artifacts_resource
+    }
+  end
+  subject { described_class.new(context).call }
+
```
miguelsorianod, line 16: Maybe a test case with a successful validation would be good

lib/3scale_toolbox/commands/plans_command/import/step.rb
miguelsorianod: I am still not fully sure I understand the purpose of the method.
`They want to get the actual metric object out of it`: Is the purpose then to find a product metric, or a backend metric depending on the provided parameters? If that's the case, why not having a `find_backend_metric` and a `find_product_metric`?

eguzki: it would be clearer if I implement  
```ruby
class Backend
  def find_metric_or_method(system_name)
  end
end

class Service
  def find_metric_or_method(system_name)
  end
end
```

?? 

Then from import plan command, when importing limits and pricing rules, we get service or backend (depending on backend system name exists or not) and then call `find_metric_or_method` method to get the metric (or method).

it would be something like this
```ruby
metric_owner = if pricing_rule['backend_system_name'].nil?
                 service
               else
                  Entities::Backend.find_by_system_name(pricing_rule['backend_system_name'])
                end
metric = metric_owner.find_metric_or_method(pricing_rule['metric_system_name'])
                              
                             
```

eguzki: and the same code in for the limits

miguelsorianod: :thinking: your alternative proposal looks good to me

spec/unit/commands/plans_command/import/validation_plan_step_spec.rb
eguzki: done

eguzki:
Found unexpected attributes in the exported plan. Fixed attribute blacklist for some entities

spec/unit/commands/plans_command/import/validation_plan_step_spec.rb
miguelsorianod: :+1:

lib/3scale_toolbox/commands/plans_command/import/step.rb
eguzki: implemented

miguelsorianod (approved):
:+1:  great work

PR #337, opened 2021-10-06
enable testing logs

eguzki:
the common expectation used in tests
```ruby
expect { subject }.to output.to_stdout
```

shadows any debug info when the command fails. Specifically, adding the env var `THREESCALE_DEBUG=1` which reports all about HTTP requests in stdout is being shadowed. By default, the debug info should go to stderr, however [this statement](https://github.com/3scale/3scale-api-ruby/blob/v1.4.0/lib/3scale/api/http_client.rb#L26) makes debug to go to stdout


miguelsorianod:
Those expectations were set to check that the toolbox outputs something, but if they are being mixed with debug logs then they are not reliable checks. If that's the case then removing them makes sense. Do we want to keep the debug logs or can we disable them for tests so this checks are reliable?

PR #340, opened 2021-11-19
backend from OAS

### LINKED ISSUE #262: Import backend from OpenAPI
It would be great to import Methods & Metrics in the Backends.

While doing an import from Swagger or OAS file, the methods & metrics should also get imported in the Backends as well along with the Product.

eguzki:
https://issues.redhat.com/browse/THREESCALE-7715

Use current Toolbox command for importing OAS and add an option to target a backend API (instead of a Product). The OAS itself won't be stored in 3scale but a backend-api, private base URL, mapping rules and methods will be created. 

`3scale import openapi -d <remote> --backend <OAS>`

Dev notes:
* The `3scale import openapi` has a new optional param: `-o json/yaml`. Since this param also affects the existing product import, I have added a report to the existing product import from OAS part that will be written to stdout only when `-o` is added. Otherwise, for backward compat, the product import from OAS still writes progress messages to stdout. 

TODO:
- [x] Tests
- [x] Doc
- [x] Validations (private host must be provided either in the OAS or from the command line)

Fixes #262


eguzki:
ready for initial review @miguelsorianod

abarrak:
Hi @eguzki,
Thanks for the much neeeded PR 🙏
Indeed `openapi import` currently is missing creating the backend, and only the service/product is created.

Also, `backend_usage` is missing and I think it needs to be added as well to complete integration between backend and product (service).

eguzki:
Thanks @abarrak 

> Also, `backend_usage` is missing and I think it needs to be added as well to complete integration between backend and product (service).

`backend_usage` attribute belongs to the product, not the backend. So, does not apply to this new functionality (although it is related of course). The question here is, how do you bind products and backends (with the `backend_usage` path) from a 3scale toolbox command? We are having discussions around that. Any input would be much appreciated.

abarrak:
@eguzki Aha, so this PR here is meant to support importing rules, methods, etc. under backend resource?

Anyway, for the `backend_usage` I'm not sure if it's possible to implement inside the import of product/backend (whichever last) or introduce a new step?

I'm writing a small plugin to toolbox, and will utilize OAS import + additional creation/update to app-plans, limits, pricing, in addition to some CMS work, and My go for the `backend_usage` is the `Entities::BackendUsage` in `3scale_toolbox/entities/backend_usage` it seems ready to use and more convenient than underlying `ruby-client' .. what do you think?

eguzki:
> Aha, so this PR here is meant to support importing rules, methods, etc. under backend resource?

Exactly. Create a 3scale backend out of a OAS doc. 

> Anyway, for the backend_usage I'm not sure if it's possible to implement inside the import of product/backend (whichever last) or introduce a new step?

The question is about which toolbox command makes sense (and what is the signature of it) to bind N backends to some product That command would need a set of N `backend_usages`, one for each N backends.

> I'm writing a small plugin to toolbox, and will utilize OAS import + additional creation/update to app-plans, limits, pricing, in addition to some CMS work, 

Cool. If you do not mind, let us have a look at it when it is finished. We can review it and add it to the core command list (maybe with some modifications) if it makes sense for somebody else.

The issue (as I see)  with the pluging you are describing is that it requires a lot of inputs. 1) the oas, 2) app plans resource. The toolbox provides two commands to do that separately:
* https://github.com/3scale/3scale_toolbox/blob/main/docs/export-import-product.md
* https://github.com/3scale/3scale_toolbox/blob/main/docs/export-import-app-plan.md

> and my go-to for the backend_usage will be mostly the Entities::BackendUsage in 3scale_toolbox/entities/backend_usage it seems ready to use and more convenient than underlying api-ruby-client .. what do you think?

To add `backend_usage`elements to a product, just use `Entities::BackendUsage`. The primitives to manage backend_usages are easy to use. The challenging part is expose a easy-to-use command for the users to manage them.

> Here's a conceptional reference on the topic: https://www.itix.fr/blog/deploy-apis-from-your-cicd-pipeline-with-the-3scale-rest-api/#backend-usage

That is the primitive I was referring just now. That is implemented in the api-ruby-client lib.

djmarinb (approved):

PR #342, opened 2021-11-24
Setup keep alive connections

eguzki:
Currently, each 3scale admin API call required to setup a new TCP/HTTP connection. For 3scale commands running multiple 3scale API calls, the toolbox spends a lot of time on TCP/HTTP connection setup. Specially true for TLS connections.

With this change, by default, the connection opened by the 3scale toolbox command (if the command requires opening a connection) will be setup to keep alive. Thus, all API calls required for the command will reuse the connection.

If, for whatever reason, the `Connection: close` behavior is desired, the PR includes a new top level option parameter:

```
--disable-keep-alive       Disable keep alive HTTP connection mode
```

### DIFF
--- a/docs/activedocs.md
+++ b/docs/activedocs.md
@@ -42,17 +42,6 @@ OPTIONS
                                        ActiveDocs
        --skip-swagger-validations      Specify it to skip validation of the
                                        Swagger specification
-
-OPTIONS FOR ACTIVEDOCS
-    -c --config-file=<value>           3scale toolbox configuration file
-                                       (default:
-                                       $HOME/.3scalerc.yaml)
-    -h --help                          show help for this command
-    -k --insecure                      Proceed and operate even for server
-                                       connections otherwise considered
-                                       insecure
-    -v --version                       Prints the version of this command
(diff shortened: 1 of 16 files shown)


docs/activedocs.md
```diff
@@ -42,17 +42,6 @@ OPTIONS
                                        ActiveDocs
        --skip-swagger-validations      Specify it to skip validation of the
                                        Swagger specification
-
```
miguelsorianod, line 45: Where this docs pending to be regenerated before this PR? I see several removals for different docs

miguelsorianod (approved):
Take a look at the comment I left and if that is expected it can be merged

docs/activedocs.md
eguzki: Those removals do not belong to this PR. They are common options, not for the specific command being documented, but for the base commands. I think the options for the base commands is not necessary to document here. I added a new option to the top level command `3scale` and I had to update all the docs for all the subcomands. So I decided to remove. 

To be clear, the base command options are still printed when you invoke help, it is just removed from the docs.

PR #357, opened 2022-10-10
ruby as the only platform

eguzki:
commands run to update
```
bundle lock --add-platform ruby
bundle lock --remove-platform powerpc64le-linux
bundle lock --remove-platform s390x-linux
bundle lock --remove-platform x86_64-linux
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -57,8 +57,6 @@ GEM
     nokogiri (1.13.6)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
-    nokogiri (1.13.6-x86_64-linux)
-      racc (~> 1.4)
     oas_parser (0.25.4)
       activesupport (>= 4.0.0)
       addressable (~> 2.3)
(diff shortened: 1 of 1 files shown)


gsaslis:
@eguzki I was wondering if there's a way we could make this "permanent" in the sense that Gemfile.lock is a generated file, and I'm wondering if we **also** need to update the process that generates it... ? 🤔 

Or is this change enough?

eguzki:
Gemnfile.lock is a generated file we manage via `bundle`.  I ran few bundle commands to change the lock file. And the lock file is being checked into git, because although being generated, it defines a state. A dep ock state.

```
bundle lock --add-platform ruby
bundle lock --remove-platform powerpc64le-linux
bundle lock --remove-platform s390x-linux
bundle lock --remove-platform x86_64-linux
```

gsaslis:
thanks @eguzki ! 

I'm good to merge this in! 👍🏼

gsaslis (approved):

